PR #10, opened 2016-08-17
BF: Handle redis errors

### LINKED ISSUE #9: Unify redis client and handle redis errors
UtapiClient currently does not listen to the redis error event, which may flood the logs when Redis
server is unavailable.
This also calls for unification of the redis client as we want the configuration to be setup in one place for easier maintenance.
Lastly, we want to remove the offline queue option, since there is no limit on the callback queue, it may grow to infinity resulting in a `RangeError`. This option can be re-introduced later after this PR on ioredis is merged, which gives the developer an option to manage the offline queue
https://github.com/luin/ioredis/pull/241

rahulreddy:
Fix #9

This commit offers a cleanup to unify the redis client usage
and add an error handler to handle redis connection errors. The option
for offline queue has been removed (for now) as it may grow the callback
queue to infinity and eventually result in a RangeError.

### DIFF
--- a/lib/UtapiClient.js
+++ b/lib/UtapiClient.js
@@ -1,4 +1,3 @@
-import Redis from 'ioredis';
 import { Logger } from 'werelogs';
 import Datastore from './Datastore';
 import { createBucket, deleteBucket, listBucket, getBucketAcl, putBucketAcl,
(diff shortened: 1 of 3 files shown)


RachedBenMustapha:
👍

LaurenSpiegel:
👍

PR #23, opened 2016-08-24
BF: Set counters to 0 on bucket creation

### LINKED ISSUE #22: set/reset global counters on bucket creation
When a push metric is received for a createBucket action, all the global counters need to be set 0 and the absolute counters (storage utilized and num. of objects) should be set to 0. This would help the following cases
- ensure all global counters are set when a bucket is created
- if a bucket is deleted and created again, global metrics from the older bucket don't "leak" into the new one.

rahulreddy:
On bucket creation, the global counters need to be set/reset to 0 indicating
the start of the bucket timeline.

Fix #22

### DIFF
--- a/lib/UtapiClient.js
+++ b/lib/UtapiClient.js
@@ -1,6 +1,6 @@
 import { Logger } from 'werelogs';
 import Datastore from './Datastore';
-import { genBucketKey } from './schema';
+import { genBucketKey, getBucketGlobalCounters } from './schema';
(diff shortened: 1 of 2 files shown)


rahulreddy:
https://github.com/scality/S3/pull/168 to check S3

nicolas2bert:
👍

rahulreddy:
I will re-open the PR after adding some tests

PR #24, opened 2016-08-29
BF: Reset counters

### LINKED ISSUE #22: set/reset global counters on bucket creation

rahulreddy:
- On bucket creation, the global counters need to be set/reset to 0 indicating
the start of the bucket timeline. Create bucket global counter is set to 1
since the action just occured.

Fix #22

- Implement in-memory backend to aid testing

### DIFF
--- a/lib/UtapiClient.js
+++ b/lib/UtapiClient.js
@@ -1,6 +1,6 @@
 import { Logger } from 'werelogs';
 import Datastore from './Datastore';
-import { genBucketKey } from './schema';
+import { genBucketKey, getBucketGlobalCounters } from './schema';
 import redisClient from '../utils/redisClient';
 
 export default class UtapiClient {
@@ -22,6 +22,17 @@ export default class UtapiClient {
                 .setClient(redisClient(config.redis, this.log));
             this.disableClient = false;
         }
+    }
+
+    /**
+    * set datastore
+    * @param {DataStore} ds - Datastore instance
+    * @return {object} current instance
+    */
+    setDataStore(ds) {
+        this.ds = ds;
+        this.disableClient = false;
+        return this;
     }
 
     /*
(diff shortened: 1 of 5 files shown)


lib/backend/Memory.js
```diff
@@ -0,0 +1,190 @@
...
+        this.db = db;
+    }
+
+    /**
+    * @param {callback} cb - callback
+    * @return {undefined}
+    */
+    exec(cb) {
+        process.nextTick(() => {
+            map(this.cmds, (item, next) => {
```
LaurenSpiegel: what is the format of this item?  what are args1, args2, args3?

lib/UtapiClient.js
```diff
@@ -46,23 +57,34 @@ export default class UtapiClient {
...
-                    });
-                    return callback(err);
-                }
-                return this.ds.zadd(genBucketKey(bucket, 'createBucket'),
-                    timestamp, count, callback);
-            });
+        const cmds = [];
+        // set global counters to 0 except for create bucket counter (set to 1)
+        // indicating start of bucket timeline
+        getBucketGlobalCounters(bucket).forEach(item => {
```
LaurenSpiegel: See https://github.com/scality/utapi/issues/25

lib/backend/Memory.js
```diff
@@ -0,0 +1,190 @@
...
+    zrangebyscore(key, min, max, cb) {
+        process.nextTick(() => {
+            if (!this.data[key]) {
+                // emulating redis-client which returns nulls
+                return cb(null, null);
+            }
+            const result = [];
+            const minScore = (min === '-inf') ? this.data[key][0][0] : min;
+            const maxScore = (min === '+inf') ?
+                this.data[key][this.data[key].length][0] : max;
```
LaurenSpiegel: -1?

lib/backend/Memory.js
```diff
@@ -0,0 +1,190 @@
...
+            this.data[key].forEach(item => {
```
LaurenSpiegel: filter?

lib/backend/Memory.js
```diff
@@ -0,0 +1,190 @@
...
+            // Sort keys by scores in the decreasing order, if scores are equal
+            // sort by their value in the decreasing order
+            cloneKeyData.sort((a, b) => {
+                if (a[0] === b[0]) {
+                    return b[1] - a[1];
+                }
+                return b[0] - a[0];
+            });
+            const result = [];
+            cloneKeyData.forEach(item => {
```
LaurenSpiegel: filter

tests/functional/testUtapiClient.js
```diff
@@ -0,0 +1,50 @@
+import assert from 'assert';
+import mapSeries from 'async/map';
```
LaurenSpiegel: misleading because there is an async map series that does the operations in order whereas map does not guaranty order.

rahulreddy: That..was a bad typo. I amended it to use named imports instead.

lib/backend/Memory.js
rahulreddy: I thought filter would not work with the linter but it does!

rahulreddy:
Addressed comments

lib/backend/Memory.js
rahulreddy: I have added JsDoc to the constructor, which defines what cmds is and what a typical example looks like.

LaurenSpiegel:
👍

nicolas2bert:
👍

PR #27, opened 2016-09-01
fix: update storage utilized for object overwrites

### LINKED ISSUE #28: Update storage utilized with delta values on object PUT overwrites
When an object is overwritten only the difference needs to be updated, as in deltaSize = newSize - oldSize

rahulreddy:
When an object is overwritten with new data the storage utilized counter should reflect only the total change in the size (storage + ( new size - old size ))

Fix #28

### DIFF
--- a/lib/UtapiClient.js
+++ b/lib/UtapiClient.js
@@ -652,23 +652,40 @@ export default class UtapiClient {
     * @param {string} bucket - bucket name
     * @param {number} timestamp - unix epoch timestamp
     * @param {number} objectSize - size of object in bytes
+    * @param {number} prevObjectSize - previous size of object in bytes if this
+    * action overwrote an existing object
     * @param {callback} [cb] - (optional) callback to call
     * @return {undefined}
     */
-    pushMetricPutObject(bucket, timestamp, objectSize, cb) {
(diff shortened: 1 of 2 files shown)


lib/UtapiClient.js
```diff
@@ -652,20 +652,23 @@ export default class UtapiClient {
...
+        const storageUtilizedDelta = objectSize - prevObjectSize;
         this.log.trace('pushing metric',
             { method: 'UtapiClient.pushMetricPutObject', bucket, timestamp });
         // update counters
         return this.ds.batch([
             ['incrby', genBucketKey(bucket, 'storageUtilizedCounter'),
-                objectSize],
+                storageUtilizedDelta],
             ['incrby', genBucketKey(bucket, 'incomingBytesCounter'),
                 objectSize],
```
alexandre-merle: you don't need to do the same here ?

alexandre-merle:
one comment, LGTM :+1:

nicolas2bert:
👍

lib/UtapiClient.js
rahulreddy: Regardless of whether the object is being overwritten, there will be bytes transferred over the network and incoming bytes is capturing just that.

PR #30, opened 2016-09-02
BF: Use zrevrangebyscore to calculate nearest values

### LINKED ISSUE #29: Incorrect calculation for absolute counters
Current logic runs into issues where no data is available during a range request and reports [0, 0] which is obviously incorrect.
Solution:
Given a [timeStart, timeEnd] use `zrevrangebyscore` command in Redis to search the nearest values to timeStart and timeEnd. For example,
```bash
zrevrangebyscore <key> <timeStart> -inf LIMIT 0 1
zrevrangebyscore <key> <timeEnd> -inf LIMIT 0 1
```
This would return the nearest values to the given time range and it works for all cases.

rahulreddy:
Current implementation returns [0,0] for absolute counters if no
data is available in the range query. This solution searches for the
nearest values to report the absolutes

Fix #29

### DIFF
--- a/lib/Buckets.js
+++ b/lib/Buckets.js
@@ -1,6 +1,6 @@
 import async from 'async';
 import { errors } from 'arsenal';
-import { getMetricFromKey, getBucketKeys } from './schema';
+import { getMetricFromKey, getBucketKeys, genBucketKey } from './schema';
 
 /**
 * Provides static methods to get bucket level metrics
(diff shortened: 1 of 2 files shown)


lib/Buckets.js
```diff
@@ -105,6 +127,45 @@ export default class Buckets {
...
+                if (item[0]) {
+                    // log error and continue
+                    log.trace('command in a batch failed to execute', {
+                        error: item[0],
+                        method: 'Buckets.getBucketMetrics',
+                    });
+                } else {
+                    let val = parseInt(item[1], 10);
+                    val = isNaN(val) ? 0 : val;
+                    if (index === 0) {
```
nicolas2bert: why not using switch here?

alexandre-merle:
LGTM :+1:

lib/Buckets.js
rahulreddy: I just prefer if..else to switch

nicolas2bert:
👍

PR #31, opened 2016-09-06
FT: use request logger for client api calls

### LINKED ISSUE #11: Instantiate a new request logger for every client api call
Currently, client api uses the module logger to log inside the client api calls. This may make debugging harder than it should be. The idea is to send requestUids to the calling method from the caller and instantiate a new request logger from that.

rahulreddy:
Fix #11


RachedBenMustapha:
👍

nicolas2bert:
👍

PR #33, opened 2016-09-06
Ft/authentication

LaurenSpiegel:
This implements an authentication and authorization step before retrieving bucket metrics.

This depends on:
1) s3: https://github.com/scality/S3/pull/189
2) arsenal: https://github.com/scality/Arsenal/pull/153

Due to batchPolicy Eval changes, still need to modify this PR to check authorization results.
Plus, for batchPolicyEval, we need to first merge:
1) S3 #204
2) Arsenal #158
3) Vault #592


router/Router.js
```diff
@@ -182,22 +183,65 @@ class Router {
...
+            log.trace('missing auth header for v4 auth');
+            return cb(errors.InvalidRequest.customizeDescription('Must ' +
+                'use Auth V4 for this request.'));
+        }
+        // resourceType will either be "buckets", "accounts" or "users"
+        const resourceType = utapiRequest.getResource();
+        const validator = utapiRequest.getValidator();
+        // specific resources will be names of buckets, accounts or users
+        const specificResources = validator.get(resourceType);
+        // TODO: once vault accepets an array of request contexts, modify this
```
rahulreddy: accepts

rahulreddy:
👍

nicolas2bert:
👍

lib/server.js
```diff
@@ -35,11 +36,14 @@ class UtapiServer {
...
         // disable nagle algorithm
         req.socket.setNoDelay();
+        const { query, path } = url.parse(req.url, true);
         const utapiRequest = new UtapiRequest()
             .setRequest(req)
             .setLog(this.logger.newRequestLogger())
             .setResponse(res)
-            .setDatastore(this.datastore);
+            .setDatastore(this.datastore)
+            .setRequestQuery(query)
```
alexandre-merle: since you use url.parse(), do you still need decodeURIComponent in those functions ?

router/Router.js
```diff
@@ -182,22 +183,65 @@ class Router {
...
+        }
+        // resourceType will either be "buckets", "accounts" or "users"
+        const resourceType = utapiRequest.getResource();
+        const validator = utapiRequest.getValidator();
+        // specific resources will be names of buckets, accounts or users
+        const specificResources = validator.get(resourceType);
+        // TODO: once vault accepts an array of request contexts, modify this
+        // to remove condition and to send array of requestContexts
+        if (specificResources.length > 1) {
+            return cb(errors.NotImplemented.customizeDescription('Users ' +
```
alexandre-merle: this message look strange, i understand users can't have multiples specific resources (agreed). But what about accounts ? Actually, this error will be raised for both user or account right ? Can be confusing :)

alexandre-merle:
little comments, LGTM otherwise :+1:

lib/server.js
LaurenSpiegel: yes.  parse just breaks it up into path and query.

router/Router.js
LaurenSpiegel: you're right.  I'll fix the error description.  My next task is to modify vault for batch policy evaluation.

router/Router.js
```diff
@@ -182,22 +183,59 @@ class Router {
...
+        const requestContexts = specificResources.map(specificResource =>
+            new policies.RequestContext(utapiRequest.getRequestHeaders(),
+            utapiRequest.getRequestQuery(), resourceType, specificResource,
+            utapiRequest.getRequesterIp(), utapiRequest.getSslEnabled(),
+            utapiRequest.getAction(), 'utapi')
+        );
+        auth.setAuthHandler(vault);
+        const requestPlusPath = utapiRequest.getRequest();
+        requestPlusPath.path = utapiRequest.getRequestPath();
+        return auth.doAuth(requestPlusPath, log, cb,
```
LaurenSpiegel: Need to handle authorization results here.

PR #37, opened 2016-09-09
FT: Push metrics: normalize timestamp to every 15 min

### LINKED ISSUE #32: Change metrics precision to second instead of millisecond
To reduce the size of datasets, metrics precision should be adjusted to the second rather than millisecond.

rahulreddy:
To reduce the number of entries in the dataset, timestamp will be
normalized to 15 minute intervals. This would mean that the metrics
will be up to date to the 15 minute interval and not millisecond.

Fix #32


LaurenSpiegel:
👍

nicolas2bert:
👍

PR #42, opened 2016-09-13
Memory backend - fix and unit tests

rahulreddy:
This PR fixes memory backend and adds a suite of unit tests to assert it's basic usage. This memory backend will in-turn be used by unit/functional tests for utapi.

### DIFF
--- a/lib/backend/Memory.js
+++ b/lib/backend/Memory.js
@@ -64,8 +64,8 @@ export default class Memory {
         assert.strictEqual(typeof value === 'string'
             || typeof value === 'number', true);
         process.nextTick(() => {
-            this.data[key] = value;
-            return cb(null, value);
+            this.data[key] = value.toString();
+            return cb(null, this.data[key]);
         });
     }
 
@@ -93,7 +93,13 @@ export default class Memory {
             if (this.data[key] === undefined) {
                 this.data[key] = 0;
             }
-            return cb(null, this.data[key]++);
+            const val = parseInt(this.data[key], 10);
+            if (Number.isNaN(val)) {
+                throw new Error('Value at key cannot be represented as a ' +
+                    'number');
+            }
+            this.data[key] = (val + 1).toString();
+            return cb(null, this.data[key]);
         });
     }
 
@@ -111,7 +117,12 @@ export default class Memory {
             if (this.data[key] === undefined) {
                 this.data[key] = 0;
             }
-            this.data[key] += num;
+            const val = parseInt(this.data[key], 10);
+            if (Number.isNaN(val)) {
+                throw new Error('Value at key cannot be represented as a ' +
+                    'number');
+            }
+            this.data[key] = (val + num).toString();
             return cb(null, this.data[key]);
         });
     }
@@ -128,7 +139,13 @@ export default class Memory {
             if (this.data[key] === undefined) {
                 this.data[key] = 0;
             }
-            return cb(null, this.data[key]--);
+            const val = parseInt(this.data[key], 10);
+            if (Number.isNaN(val)) {
+                throw new Error('Value at key cannot be represented as a ' +
+                    'number');
+            }
+            this.data[key] = (val - 1).toString();
+            return cb(null, this.data[key]);
         });
     }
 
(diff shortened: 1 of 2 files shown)


rahulreddy:
This PR relies on commits from previous PR, please review only the last 2 commits
- fix: memory backend to aid unit/functional testing
- test: memory backend implementation

lib/backend/Memory.js
```diff
@@ -146,7 +163,12 @@ export default class Memory {
             if (this.data[key] === undefined) {
                 this.data[key] = 0;
             }
-            this.data[key] -= num;
+            const val = parseInt(this.data[key], 10);
+            if (Number.isNaN(val)) {
+                throw new Error('Value at key cannot be represented as a ' +
```
michael-zapata: Why throw when we can use the callback to send back an error?

lib/backend/Memory.js
```diff
@@ -93,7 +93,13 @@ export default class Memory {
             if (this.data[key] === undefined) {
                 this.data[key] = 0;
             }
-            return cb(null, this.data[key]++);
+            const val = parseInt(this.data[key], 10);
+            if (Number.isNaN(val)) {
+                throw new Error('Value at key cannot be represented as a ' +
```
michael-zapata: Why throw when we can use the callback to send back an error?

lib/backend/Memory.js
```diff
@@ -144,25 +177,31 @@ export default class Memory {
     * Store value by score in a sorted set
     * @param {string} key - data key
     * @param {number} score - data score
-    * @param {number} value - data value
+    * @param {string} value - data value
```
michael-zapata: Can be a number as well if we can trust your assertions.

lib/backend/Memory.js
```diff
@@ -83,11 +88,18 @@ export default class Memory {
...
     incr(key, cb) {
+        assert.strictEqual(typeof key, 'string');
         process.nextTick(() => {
             if (this.data[key] === undefined) {
                 this.data[key] = 0;
             }
-            return cb(null, this.data[key]++);
+            const val = parseInt(this.data[key], 10);
+            if (Number.isNaN(val)) {
+                throw new Error('Value at key cannot be represented as a ' +
```
michael-zapata: Why throw when we can use the callback to send back an error?

lib/backend/Memory.js
rahulreddy: Added number as the type as well.

lib/backend/Memory.js
rahulreddy: The list metrics api call ignores error from a command, logs them and moves on to the next. So if errors are sent in callbacks we would never catch them in the tests.  Since this Memory backend is intended purely for testing (unit/functional) I have decided to purposefully throw errors which would stop the test execution. It's the same reasoning for assertions, it's so that we would catch them in the tests.

lib/backend/Memory.js
```diff
@@ -272,11 +295,12 @@ export default class Memory {
             const maxScore = (min === '+inf') ?
                 this.data[key][this.data[key].length][0] : max;
             let removeCount = 0;
-            this.data[key].forEach((item, index) => {
+            this.data[key] = this.data[key].filter(item => {
                 if (item[0] >= minScore && item[0] <= maxScore) {
-                    this.data[key].splice(index, 1);
                     removeCount++;
+                    return false;
```
LaurenSpiegel: Instead of having a side effect in your filter you could check the lengths before and after. But up to you.

LaurenSpiegel:
👍

lib/backend/Memory.js
rahulreddy: Yeah that's better, I made the changes.

lib/backend/Memory.js
```diff
@@ -239,14 +294,14 @@ export default class Memory {
...
             const maxScore = (min === '+inf') ?
                 this.data[key][this.data[key].length][0] : max;
-            let removeCount = 0;
-            this.data[key].forEach((item, index) => {
+            const oldLen = this.data[key].length;
+            this.data[key] = this.data[key].filter(item => {
                 if (item[0] >= minScore && item[0] <= maxScore) {
-                    this.data[key].splice(index, 1);
-                    removeCount++;
+                    return false;
```
michael-zapata: Since you only have a conditional in that filter, no need to keep the returning boolean clutter, you can just use lambda syntax.

rahulreddy: Updated.

rahulreddy:
@michael-zapata Merging this before you could resolve as we need this for the builds.

PR #45, opened 2016-09-16
fix: operations counters

### LINKED ISSUE #44: operations count does not increment on multiple actions
For example, doing 5 object PUTs we expect the operations count in the response to be 5, instead it is 1.

rahulreddy:
PR #37 of reducing the time precision introduced a bug which made
calculating operations incorrectly. As the code no longer adds multiple
members for the same timestamp (counters are updated for the normalized
timestamp), the solution has been simplified as follows

- All keys which hold operations counters (put object, create bucket etc.)
  use string keys instead of sorted sets

- INCR cmd is issued for the normalized timestamp on key on every push
  metric

- Total number of operations between start and end timestamps is the sum
  of counts held by the keys that fall between the two timestamps

- As sorted sets are no longer used for operations, there is no need for
  global counters for these keys, hence they are removed

- The only metrics which use sorted sets are the ones holding bucket state
  data - storage utilized and number of objects. Sorted sets are the best
  candidates for finding the nearest neighbor values

Fix #44

### DIFF
--- a/lib/Buckets.js
+++ b/lib/Buckets.js
@@ -1,7 +1,6 @@
 import async from 'async';
 import { errors } from 'arsenal';
 import { getMetricFromKey, getBucketKeys, genBucketStateKey } from './schema';
-import getTimespan from '../utils/getTimespan';
 
 /**
 * Provides static methods to get bucket level metrics
@@ -35,19 +34,19 @@ export default class Buckets {
     }
 
     /**
-    * Returns a list of timestamps incremented by a day starting from start
-    * to end
+    * Returns a list of timestamps incremented by 15 min. from start timestamp
+    * to end timestamp
     * @param {number} start - start timestamp
     * @param {number} end - end timestamp
     * @return {number[]} range - array of timestamps
     */
-    static getTimespanRange(start, end) {
+    static getTimestampRange(start, end) {
         const res = [];
         let last = start;
-        while (last !== end) {
+        while (last < end) {
             res.push(last);
             const d = new Date(last);
-            last = d.setDate(d.getDate() + 1);
+            last = d.setMinutes(d.getMinutes() + 15);
         }
         res.push(end);
         return res;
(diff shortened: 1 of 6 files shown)


rahulreddy:
This PR has commits from other PRs. Please review **fix: operations counters** commit only.

lib/Buckets.js
```diff
@@ -150,15 +137,13 @@ export default class Buckets {
...
                     's3:HeadBucket': 0,
                     's3:HeadObject': 0,
                 },
             };
 
-            // last 8 are results of storageUtilized, numberOfObjects,
-            // incomingBytes and outgoingBytes
-            const absolutes = res.splice(-8);
+            // last 4 are results of storageUtilized, numberOfObjects,
+            const absolutes = res.splice(-4);
```
LaurenSpiegel: I would do two slices.

rahulreddy: I made the change, I know that it enhances code readability.

lib/Buckets.js
```diff
@@ -212,12 +183,15 @@ export default class Buckets {
...
-                } else if (Array.isArray(item[1])) {
-                    // convert strings to numbers for comparision
+                } else {
                     const m = getMetricFromKey(key, bucket);
-                    // count is enough as it represents the delta of the
-                    // metric between the two timestamps
-                    bucketRes.operations[`s3:${m}`] += item[1].length || 0;
+                    let count = parseInt(item[1], 10);
+                    count = Number.isNaN(count) ? 0 : count;
+                    if (m === 'incomingBytes' || m === 'outgoingBytes') {
```
LaurenSpiegel: why aren't these just called s3:incomingBytes as well?

lib/UtapiClient.js
```diff
@@ -861,69 +656,32 @@ export default class UtapiClient {
...
             }
+            cmds.push(
+                ['zremrangebyscore',
+                    genBucketStateKey(bucket, 'storageUtilized'),
+                    timestamp, timestamp],
+                ['zadd', genBucketStateKey(bucket, 'storageUtilized'),
+                    timestamp, actionCounter]);
 
-            // incoming bytes counter
+            // number of objects counter
```
LaurenSpiegel: Is this correct?  You have number of objects counter above.

lib/Buckets.js
rahulreddy: because that's the key we are using in our responses. The `s3:` prefix was added only for the api operations list (s3:putObject, s3:getObject etc.)

lib/UtapiClient.js
rahulreddy: incoming bytes has been removed, so I updated the comment. Storage utilized is line 650

lib/UtapiClient.js
```diff
@@ -409,67 +319,40 @@ export default class UtapiClient {
...
             ['incr', genBucketCounter(bucket, 'numberOfObjectsCounter')],
-            ['incr',
-                genBucketCounter(bucket, 'completeMultipartUploadCounter')],
+            ['incr', genBucketKey(bucket, 'completeMultipartUpload',
+                timestamp)],
         ], (err, results) => {
-            // number of objects counter
             if (err) {
                 log.error('error incrementing counter for push metric', {
                     method: 'UtapiClient.pushMetricPutObject',
```
alexandre-merle: pushMetricCompleteMultipartUpload

rahulreddy: Fixed.

PR #48, opened 2016-09-19
fix: utapi config overrides

### LINKED ISSUE #47: Utapi does not set https, vaultd config from S3
Utapi config needs to be overridden by the config passed from S3.

rahulreddy:
Utapi's config read from config.json is overridden by the config passed from
S3. This commit fixes
 - https config passed down from S3
 - vaultd config to use the overridden config from S3

Fix #47

### DIFF
--- a/lib/Vault.js
+++ b/lib/Vault.js
@@ -1,26 +1,22 @@
 import vaultclient from 'vaultclient';
 
-import config from './Config';
-
-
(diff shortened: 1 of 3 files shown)


rahulreddy:
end to end passed http://ci.ironmann.io/gh/scality/Integration/3769

PR #49, opened 2016-09-21
FT: compile using babel

rahulreddy:
Included commits

### ft: compile js to dist dir using babel
  As utapi is a module that will be required by other packages, the code
will be compiled by babel to the dist directory for distribution.

### 	ft: ignore dist folder
babel compiles from src to dist


package.json
```diff
@@ -31,6 +32,7 @@
     "mocha": "^3.0.2"
   },
   "scripts": {
+    "postinstall": "./node_modules/.bin/babel --ignore node_modules src/ --out-dir dist",
```
michael-zapata: You can just use `babel` instead, npm will infer that folder.

package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "arsenal": "scality/Arsenal#rel/6.2",
     "async": "^2.0.1",
+    "babel-cli": "^6.14.0",
```
michael-zapata: Could we freeze the version instead?

package.json
rahulreddy: Ah, I wasn't sure if npm would pick that up.

package.json
rahulreddy: Sure

rahulreddy:
@michael-zapata I have addressed your comments.

Issue #55, opened 2016-10-03
Start time is later than end time


nicolas2bert:
Should return InvalidParameterValue if start time is later than end time.

PR #56, opened 2016-10-03
valid time range check Fix issue #55

nicolas2bert:
Fix https://github.com/scality/utapi/issues/55


rahulreddy:
Can you change the commit message to something like `valid time range check`

PR #63, opened 2016-10-20
ft: utapi healthcheck route

rahulreddy:
This commit adds a healthcheck route bypassing the traditional router.
The route responds back with 200 if the server has a successful connection to
Redis and 500 if not. Additionally, the router checks IP address from the request
and denies any IP not whitelisted.

### DIFF
--- a/src/lib/Config.js
+++ b/src/lib/Config.js
@@ -58,6 +58,20 @@ class Config {
             }
         }
 
+        this.healthChecks = { allowFrom: ['127.0.0.1/8', '::1'] };
+        if (config.healthChecks && config.healthChecks.allowFrom) {
+            assert(Array.isArray(config.healthChecks.allowFrom),
+                'config: invalid healthcheck configuration. allowFrom must ' +
+                'be an array');
+            config.healthChecks.allowFrom.forEach(item => {
+                assert(typeof item === 'string',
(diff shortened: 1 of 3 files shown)


rahulreddy:
I made a hacky implementation at the moment.
The cleaner implementation requires a new router to setup admin routes which would not follow the conventional Utapi request spec. and auth model. The test suite setup needs to change as well to start the server as needed. These would be addressed with a PR against master. For now, we have this implementation.

rahulreddy:
@ironman-machine ...

src/lib/Config.js
```diff
@@ -58,6 +58,20 @@ class Config {
             }
         }
 
+        this.healthChecks = { allowFrom: ['127.0.0.1/8', '::1'] };
+        if (config.healthChecks && config.healthChecks.allowFrom) {
+            assert(config.healthChecks.allowFrom instanceof Array,
```
LaurenSpiegel: I don't know if using Array.isArray rather than instanceof matters in node but it is the preferred method at least for web.

rahulreddy: I was reading up on it and it was because browsers run into weird scoping issues due to the global `window` object. Either way, I will make the change as it's a good practice.

rahulreddy:
@ironman-machine again

PR #64, opened 2016-10-21
Add multiObjectDelete action

LaurenSpiegel:
This should be merged with https://github.com/scality/Integration/pull/284

### DIFF
--- a/src/lib/UtapiClient.js
+++ b/src/lib/UtapiClient.js
@@ -453,33 +453,38 @@ export default class UtapiClient {
     }
 
     /**
-    * Updates counter for DeleteObject action on an object of Bucket resource.
+    * Updates counter for DeleteObject or MultiObjectDelete action
     * @param {string} reqUid - Request Unique Identifier
     * @param {string} bucket - bucket name
-    * @param {number} objectSize - size of the object
+    * @param {number} objectSize - size of the object(s)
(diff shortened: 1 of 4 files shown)


michael-zapata:
Question: Shouldn't the metric count as multiple delete instead? Given that you can erase between 1 and 1000 objects in one go, we lose a lot of information only incrementing once.

src/lib/UtapiClient.js
```diff
@@ -453,33 +453,38 @@ export default class UtapiClient {
...
     * @return {undefined}
     */
-    pushMetricDeleteObject(reqUid, bucket, objectSize, cb) {
+    _genericPushMetricDeleteObject(reqUid, bucket, objectSize,
+        numOfObjects, cb) {
         const callback = cb || this._noop;
         if (this.disableClient) {
             return callback();
         }
+        const bucketAction = numOfObjects === 1 ? 'deleteObject'
```
michael-zapata: Why not use a boolean instead to differ between the two operations?

LaurenSpiegel:
@michael-zapata we're implementing a different api call.  We don't lose the info because we decrement the number of objects and size.

src/lib/UtapiClient.js
michael-zapata: Ok, I did not see that these strings have purpose afterwards.

PR #71, opened 2016-12-15
FT: Update storage utilized on part overwrite

bennettbuchanan:
If the upload part request is an overwrite of an existing object, use the `storageUtilizedDelta` to calculate the correct storage utilized value.

See associated PRs: 
https://github.com/scality/S3/pull/472
https://github.com/scality/Integration/pull/384

### DIFF
--- a/src/lib/UtapiClient.js
+++ b/src/lib/UtapiClient.js
@@ -331,15 +331,17 @@ export default class UtapiClient {
     * @return {undefined}
     */
     _pushMetricUploadPart(params, timestamp, action, log, callback) {
-        this._checkProperties(params, ['newByteLength']);
-        const { newByteLength } = params;
+        this._checkProperties(params, ['newByteLength', 'oldByteLength']);
         this._logMetric(params, '_pushMetricUploadPart', timestamp, log);
         const cmds = [];
+        const { newByteLength, oldByteLength } = params;
+        const oldObjSize = oldByteLength === null ? 0 : oldByteLength;
+        const storageUtilizedDelta = newByteLength - oldObjSize;
         const paramsArr = this._getParamsArr(params);
         paramsArr.forEach(p => {
             cmds.push(
                 ['incrby', generateCounter(p, 'storageUtilizedCounter'),
-                    newByteLength],
+                    storageUtilizedDelta],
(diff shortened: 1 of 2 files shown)


bennettbuchanan:
end-to-end passed here: http://ci.ironmann.io/gh/scality/Integration/7480

src/lib/UtapiClient.js
```diff
@@ -256,22 +256,32 @@ export default class UtapiClient {
...
         if (this.disableClient) {
             return callback();
         }
         const log = this.log.newRequestLoggerFromSerializedUids(reqUid);
         const timestamp = UtapiClient.getNormalizedTimestamp();
+
+        let oldObjSize = parseInt(prevObjectSize, 10);
+        oldObjSize = isNaN(oldObjSize) ? 0 : oldObjSize;
+        let newObjSize = parseInt(objectSize, 10);
+        newObjSize = isNaN(newObjSize) ? 0 : newObjSize;
```
LaureVergeron: In both these situations, what are the contexts when parseInt could return NaN? Shouldn't we be logging something then (if a wrong value was passed, there is probably some interesting context associated with it?), and maybe stop function execution here rather than assume a size of zero and go along?

bennettbuchanan: The addition of logs for incorrect value types is a good direction, and one that we are addressing in the https://github.com/scality/utapi/pull/72 refactor of Utapi. With this PR, we are instead enforcing expected types for push metric actions using assertion statements.

`prevObjectSize` can be `null` if the `uploadPart` request is not an overwrite of pre-existing data. In this case, `oldObjSize` should be `0` to create the correct `storageUtilizedDelta`. 

As for `objectSize`, I am following the pattern of `pushMetricPutObject` in which the same situation applies—I assumed there may be a case I do not know of. That value comes from `request.parsedContentLength` in the objectPutPart API.

LaureVergeron:
LGTM now

bennettbuchanan:
http://ci.ironmann.io/gh/scality/Integration/9336

bennettbuchanan:
e2e passed

rahulreddy:
Please rebase this and run and end to end without any branch deps.

bennettbuchanan:
http://ci.ironmann.io/gh/scality/Integration/9475

PR #72, opened 2016-12-16
Refactor/utapi client

rahulreddy:
This commit refactors all the older methods for pushing metrics to expose one
single method for pushing all metrics. This will help accommodating other
granularities (account, user, service) in the upcoming PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,44 @@
 
 Service Utilization API for tracking resource usage and metrics reporting
 
+## Client
+
+The module exposes a client, named UtapiClient. Projects can use this client to
+push metrics directly to the underlying datastore(Redis) without the need of an
+extra HTTP request to Utapi.
+
+```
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -7,6 +7,43 @@
...
+
+const config = {
+    redis: {
+        host: '127.0.0.1',
+        port: 6379
+    }
+}
+const c = new UtapiClient(config);
+
+c.pushMetric('createBucket', '3d534b1511e5630e68f0', { bucket: 'demo' });
```
electrachong: Only comment is you can add a comment to explain the parameters if you don't think the people reading the README.md will be familiar with Utapi already. Mostly it is the reqUid that is not clear at first glance what it represents.

bennettbuchanan: Thanks! We will open another PR to expand this README.

PR #74, opened 2016-12-31
FT: Add account metrics to UtapiClient

bennettbuchanan:
Refactors the schema keys to accommodate additional metric granularities and updates `UtapiClient` to push account level metrics.

### DIFF
--- a/src/lib/UtapiClient.js
+++ b/src/lib/UtapiClient.js
@@ -1,36 +1,43 @@
 import assert from 'assert';
+import async from 'async';
 import { Logger } from 'werelogs';
 import Datastore from './Datastore';
-import { genBucketKey, genBucketCounter, getBucketCounters, genBucketStateKey }
+import { generateKey, generateCounter, getCounters, generateStateKey }
     from './schema';
 import { errors } from 'arsenal';
 import redisClient from '../utils/redisClient';
 
 const methods = {
     createBucket: '_pushMetricCreateBucket',
-    deleteBucket: '_pushMetricDeleteBucket',
-    listBucket: '_pushMetricListBucket',
-    getBucketAcl: '_pushMetricGetBucketAcl',
-    putBucketAcl: '_pushMetricPutBucketAcl',
-    putBucketWebsite: '_pushMetricPutBucketWebsite',
-    getBucketWebsite: '_pushMetricGetBucketWebsite',
-    deleteBucketWebsite: '_pushMetricDeleteBucketWebsite',
+    deleteBucket: '_genericPushMetric',
+    listBucket: '_genericPushMetric',
+    getBucketAcl: '_genericPushMetric',
+    putBucketAcl: '_genericPushMetric',
+    putBucketWebsite: '_genericPushMetric',
+    getBucketWebsite: '_genericPushMetric',
+    deleteBucketWebsite: '_genericPushMetric',
     uploadPart: '_pushMetricUploadPart',
-    initiateMultipartUpload: '_pushMetricInitiateMultipartUpload',
+    initiateMultipartUpload: '_genericPushMetric',
     completeMultipartUpload: '_pushMetricCompleteMultipartUpload',
     listMultipartUploads: '_pushMetricListBucketMultipartUploads',
-    listMultipartUploadParts: '_pushMetricListMultipartUploadParts',
-    abortMultipartUpload: '_pushMetricAbortMultipartUpload',
-    deleteObject: '_pushMetricDeleteObject',
-    multiObjectDelete: '_pushMetricMultiObjectDelete',
+    listMultipartUploadParts: '_genericPushMetric',
+    abortMultipartUpload: '_genericPushMetric',
+    deleteObject: '_genericPushMetricDeleteObject',
+    multiObjectDelete: '_genericPushMetricDeleteObject',
     getObject: '_pushMetricGetObject',
-    getObjectAcl: '_pushMetricGetObjectAcl',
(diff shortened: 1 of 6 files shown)


src/lib/UtapiClient.js
```diff
@@ -80,27 +84,30 @@ export default class UtapiClient {
...
 
    /**
-    * Check the types of optional `params` object properties. This enforces
-    * object properties for particular push metric calls and property types.
+    * Check the types of `params` object properties. This enforces object
+    * properties for particular push metric calls and enforces the existence of
+    * at least one metric type (e.g., 'bucket' or 'account').
     * @param {object} params - params object with metric data
-    * @param {string} params.bucket - bucket name
+    * @param {string} params.bucket - (optional) bucket name
```
electrachong: you can put in brackets around `params.bucket` to indicate it's optional
http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values

same for the other optional params

src/lib/UtapiClient.js
```diff
@@ -113,13 +120,55 @@ export default class UtapiClient {
...
+     * Creates an array of parameter objects for each metric type. The number
+     * of objects in the array will be the number of metric types included in
+     * the `params` object.
+     * @param {object} params - params object with metric data
+     * @return {object []} arr - array of parameter objects for push metric call
+     */
+    _getParamsArr(params) {
+        const arr = [];
+        const types = metricTypes.filter(type => type in params);
+        types.forEach(type => {
```
electrachong: looks like you can use a .map here since you are creating a new array with items that correspond to items of the types array

src/lib/UtapiClient.js
```diff
@@ -113,13 +120,55 @@ export default class UtapiClient {
...
+            const typeObj = {};
```
electrachong: shorthand:
           `const typeObj = { type: params[type] };`

src/lib/UtapiClient.js
```diff
@@ -113,13 +120,55 @@ export default class UtapiClient {
...
     * Generic method exposed by the client to push a metric with some values.
-    * `params` can be expanded to provide metrics for other granularities
-    * (e.g. service, account, user).
+    * `params` can be expanded to provide metrics for metric granularities
+    * (e.g. 'bucket', 'account').
     * @param {string} metric - metric to be published
     * @param {string} reqUid - Request Unique Identifier
     * @param {object} params - params object with metric data
-    * @param {string} params.bucket - bucket name
+    * @param {string} params.bucket - (optional) bucket name
```
electrachong: same comment about the optional params for jsdoc comments

src/lib/UtapiClient.js
```diff
@@ -131,82 +180,65 @@ export default class UtapiClient {
     * @return {object} this - current instance
     */
     pushMetric(metric, reqUid, params, cb) {
+        assert(methods[metric], `${metric} metric is not handled by Utapi`);
```
electrachong: I was wondering about this, nice that methods that are not yet supported will get an error response.
Does this mean I can uncomment pushmetric calls in my s3 pull requests for bucketCors actions, and just update Utapi/Integration later?

bennettbuchanan: Pushing a metric that Utapi cannot handle will cause the API's callback to never be called, as it throws an `AssertionError`. If you were testing it, then this assertion would be handy for indicating the problem—it will tell you which metric is unable to be pushed. The same applies for object properties expected for each API's metric, and the property types. Utapi is user-friendly! 😺

src/lib/UtapiClient.js
bennettbuchanan: Wouldn't `typeObj` end up as `{ type: 'foo' }` instead of, for example, `{ bucket: 'foo' }`? Maybe I'm missing something.

src/lib/UtapiClient.js
bennettbuchanan: Thank you. That is a cleaner approach.

src/lib/UtapiClient.js
bennettbuchanan: Thank you for pointing that out. Updated.

src/lib/UtapiClient.js
electrachong: Oh, you're right. Sloppy reading on my behalf

tests/unit/testUtapiClient.js
```diff
@@ -9,18 +9,79 @@ const memoryBackend = new MemoryBackend();
...
+    bucket: 'foo-bucket',
+    accountId: 'foo-account',
+};
+
+// Get prefix values to construct the expected Redis schema keys
+function getPrefixValues(timestamp) {
+    const arr = [];
+    Object.keys(metricTypes).forEach(metric => {
+        if (metricTypes[metric] === undefined) {
+            return;
```
electrachong: for my understanding, you can't return from a `forEach` loop:
> There is no way to stop or break a forEach() loop other than by throwing an exception. If you need such behavior, the forEach() method is the wrong tool, use a plain loop instead. If you are testing the array elements for a predicate and need a Boolean return value, you can use every() or some() instead. If available, the new methods find() or findIndex() can be used for early termination upon true predicates as well.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach

bennettbuchanan: I was using that to test undefined object values. I've removed it.

bennettbuchanan:
@ironman-machine

PR #75, opened 2017-01-04
FT: Add account metrics to ListMetrics

bennettbuchanan:
Refactors ListMetrics and server routes to accommodate additional metric granularities.


bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine

bennettbuchanan:
SCALITY_INTEGRATION_BRANCH=test/add_utapi_account_metrics
@ironman-machine try

bennettbuchanan:
SCALITY_INTEGRATION_BRANCH=test/add_utapi_account_metrics
@ironman-machine try

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
SCALITY_INTEGRATION_BRANCH=test/add_utapi_account_metrics
SCALITY_S3_BRANCH=ft/add_list_metrics_account_level
@ironman-machine try

src/handlers/metricsHandlers.js
```diff
@@ -0,0 +1,35 @@
+import ListMetrics from '../lib/ListMetrics';
+
+export class BucketsHandler {
```
rahulreddy: This could be one class `ListMetricsHandler`, you can pass the type `buckets/accounts` to the constructor to instantiate appropriate metrics.
The filename needs to reflect the classname for readability - `ListMetricsHandler.js`

src/lib/ListMetrics.js
```diff
@@ -3,33 +3,101 @@ import { errors } from 'arsenal';
...
+            type = 'accountId';
+        }
+        const obj = {};
+        obj[type] = value;
+        return obj;
+    }
+
+    // Create the metric response object for a given metric.
+    _getMetricRes(value, start, end) {
+        const metricRes = {
```
rahulreddy: Let's move this to a JSON file. The benefit would be that it would be self-validating for syntax, ListMetrics would be smaller, the JSON would be loaded only once in the memory.  Make sure you make a deep copy of the JSON before using it.

src/lib/ListMetrics.js
```diff
@@ -3,33 +3,101 @@ import { errors } from 'arsenal';
...
-    * @param {Buckets~bucketsMetricsCb} cb - callback
+    * @param {ListMetrics~bucketsMetricsCb} cb - callback
     * @return {undefined}
     */
-    static getBucketsMetrics(utapiRequest, cb) {
+    getTypesMetrics(utapiRequest, cb) {
         const log = utapiRequest.getLog();
         const validator = utapiRequest.getValidator();
-        const buckets = validator.get('buckets');
+        const metric = validator.get(this.metric);
```
rahulreddy: Let's call the list of accounts or buckets as resources and use the language consistently across the file. This will help not to confuse it with metric type.

src/lib/ListMetrics.js
```diff
@@ -40,7 +108,7 @@ export default class Buckets {
     * @param {number} end - end timestamp
     * @return {number[]} range - array of timestamps
     */
-    static getTimestampRange(start, end) {
+    _getTimestampRange(start, end) {
```
rahulreddy: Why was the static keyword removed?

src/lib/ListMetrics.js
```diff
@@ -97,56 +166,23 @@ export default class Buckets {
...
+        const metricKeys = [].concat.apply([], timestampRange.map(
+            i => getKeys(obj, i)));
+        const cmds = metricKeys.map(item => ['get', item]);
         cmds.push(storageUtilizedStart, storageUtilizedEnd,
             numberOfObjectsStart, numberOfObjectsEnd);
 
         datastore.batch(cmds, (err, res) => {
             if (err) {
-                log.trace('error occurred while getting bucket metrics', {
+                log.trace('error occurred while getting metric metrics', {
```
rahulreddy: error occurred while getting ~~metric~~ metrics

tests/unit/testMetrics.js
```diff
@@ -0,0 +1,204 @@
...
+};
+const metricLevels = {
+    bucket: 'buckets',
+    accountId: 'accounts',
+};
+datastore.setClient(memBackend);
+
+// Create the metric response object for a given metric.
+function getMetricRes(type) {
+    const metricRes = {
```
rahulreddy: Once you make this a JSON, it can be imported directly here.

src/handlers/metricsHandlers.js
bennettbuchanan: Updated. Thanks!

src/lib/ListMetrics.js
bennettbuchanan: The method which uses this is no longer a `static` class.

src/lib/ListMetrics.js
bennettbuchanan: Updated

tests/unit/testMetrics.js
bennettbuchanan: Updated.

src/lib/ListMetrics.js
bennettbuchanan: Updated.

src/lib/ListMetrics.js
bennettbuchanan: Updated.

src/lib/ListMetrics.js
```diff
@@ -1,35 +1,73 @@
...
+
+    // Create the metric response object for a given metric.
+    _getMetricResponse(resource, start, end) {
+        const metricResponse = Object.assign({}, metricResponseJSON);
+        metricResponse.timeRange = [start, end];
+        const map = {
+            buckets: 'bucketName',
+            accounts: 'accountId',
+        };
+        metricResponse[map[this.metric]] = resource;
```
dora-korpar: I'm curious what 'map' is doing here. Wouldn't it be enough to just say metricResponse[this.metric] = resource?

bennettbuchanan: It is getting the associated metric key (`bucketName` or `accountId`) from the `map` object above. 

`metricResponse` should have one of those keys for the associated metric type's (buckets or accounts) resource. There should not be a key `buckets` or `accounts` which is what `this.metric` would be.

bennettbuchanan: Maybe `map` is the wrong name for this object.

dora-korpar: oooh yeah, I thought it was calling the map method in a fancy new way. Maybe a name like 'metricMap' or sth would make it a little clearer

bennettbuchanan:
end-to-end passed here: http://ci.ironmann.io/gh/scality/Integration/8518

src/lib/ListMetrics.js
bennettbuchanan: Thanks, I've updated this for clarity. 😄

bennettbuchanan:
@ironman-machine go

bennettbuchanan:
end-to-end passed

PR #77, opened 2017-01-17
REFACTOR: Eliminate loops

bennettbuchanan:
Remove unnecessary loops in params array construction in UtapiClient by creating the object directly. This is possible because any irrelevant properties are already ignored in the pushMetric calls.

Also, remove two unnecessary `find` methods.

### DIFF
--- a/src/lib/UtapiClient.js
+++ b/src/lib/UtapiClient.js
@@ -59,8 +59,15 @@ export default class UtapiClient {
                 .setClient(redisClient(config.redis, this.log));
             this.disableClient = false;
         }
+        // Use metric levels included in the config, or use all metrics.
         if (config && config.metrics) {
+            assert(Array.isArray(config.metrics), '`metrics` property of ' +
+            'Utapi configuration must be an array');
+            assert(config.metrics.length !== 0, '`metrics` property of Utapi ' +
+            'configuration must contain at least one metric');
             this.metrics = config.metrics;
+        } else {
+            this.metrics = ['buckets', 'accounts'];
         }
     }
 
@@ -130,18 +137,15 @@ export default class UtapiClient {
      * @return {undefined}
      */
     _checkMetricTypes(params) {
+        assert(this.metrics.some(level =>
+            metricObj[level] in params), 'Must include a metric level');
         // Object of metric types and their associated property names
-        if (this.metrics) {
-            this.metrics.forEach(level => {
-                const propName = metricObj[level];
-                assert(typeof params[propName] === 'string' ||
-                    params[propName] === undefined,
-                    `${propName} must be a string`);
-            });
-        } else {
-            assert(Object.keys(metricObj).some(level =>
-                metricObj[level] in params), 'Must include a metric level');
-        }
+        this.metrics.forEach(level => {
+            const propName = metricObj[level];
+            assert(typeof params[propName] === 'string' ||
+                params[propName] === undefined,
+                `${propName} must be a string`);
+        });
     }
 
     /**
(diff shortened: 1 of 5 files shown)


bennettbuchanan:
@ironman-machine go

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
e2e: http://ci.ironmann.io/gh/scality/Integration/9206

src/lib/UtapiClient.js
```diff
@@ -169,33 +167,26 @@ export default class UtapiClient {
...
         // Only push metric levels defined in the config, otherwise push any
         // levels that are passed in the object
-        const levels = this.metrics ? this.metrics : Object.keys(metricObj);
-        const props = [];
-        for (let i = 0; i < levels.length; i++) {
-            const prop = metricObj[levels[i]];
-            if (params[prop] !== undefined) {
-                props.push(metricObj[levels[i]]);
-            }
+        const levels = this.metrics ? this.metrics : metricLevels;
```
rahulreddy: See my comment above, this can be taken care of in the constructor, keep it DRY.

src/lib/UtapiClient.js
```diff
@@ -169,33 +167,26 @@ export default class UtapiClient {
...
+        if (params.bucket && levels.indexOf('buckets') >= 0) {
```
rahulreddy: You can use `Array.includes` prop. here. It just wraps indexOf and is more readable.
```js
levels.includes('buckets')
```

src/lib/schema.js
```diff
@@ -45,16 +45,11 @@ const keys = {
...
-        bucket: 'buckets',
-        accountId: 'accounts',
-    };
-    const metricType = Object.keys(params).find(k => k in map);
-    const level = map[metricType];
-    const id = params[metricType];
-    const prefix = timestamp === undefined ? `s3:${level}:${id}:` :
-        `s3:${level}:${timestamp}:${id}:`;
+    const { bucket, accountId } = params;
+    const level = bucket ? 'buckets' : 'accounts';
```
rahulreddy: Same, DRY.

src/lib/UtapiClient.js
```diff
@@ -169,33 +167,26 @@ export default class UtapiClient {
      */
     _getParamsArr(params) {
         this._checkMetricTypes(params);
+        const props = [];
+        const { byteLength, newByteLength, oldByteLength, numberOfObjects } =
+            params;
+        const metricData = {
```
electrachong: you can probably just do `const metricData = Object.assign({}, params)` right? Is there an advantage to doing it this way?

src/lib/UtapiClient.js
bennettbuchanan: Using this method on an array throws TypeError.

src/lib/UtapiClient.js
bennettbuchanan: We cannot pass `params` here because this function creates isolated objects for each metric level ('buckets', 'accounts', etc.). The `params` object can contain both levels (for example, how S3 passes an object with 'bucket' and 'accountId' as properties). I've now added a more explicit comment in the JSDoc here.

electrachong: hmm OK, I think you mean `params` has more properties than the ones listed here, right? In other words

src/lib/UtapiClient.js
electrachong: That's weird? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes do you mean because `buckets` is an array?

src/lib/UtapiClient.js
bennettbuchanan: Yes, *sensitive* properties. 😄

src/lib/UtapiClient.js
bennettbuchanan: With node v4.5, at least it does.

bennettbuchanan:
http://ci.ironmann.io/gh/scality/Integration/9262

bennettbuchanan:
e2e after squash: http://ci.ironmann.io/gh/scality/Integration/9289

bennettbuchanan:
e2e with Integration: http://ci.ironmann.io/gh/scality/Integration/9293

PR #78, opened 2017-01-19
Ft/transactions

rahulreddy:
This PR includes the following commits

- **ft: use multi/exec instead of vanilla pipeline**

  This commit ensures that a batch of commands are isolated during execution.

- **ft: send all metrics to redis in the same transaction**

  This commit ensures that commands for all metric levels buckets, accounts,
users, service etc. are pushed in the same transaction. This guarantees
atomicity and saves on the round trips to the redis server.

### DIFF
--- a/src/lib/UtapiClient.js
+++ b/src/lib/UtapiClient.js
@@ -1,5 +1,4 @@
 import assert from 'assert';
-import async from 'async';
 import { Logger } from 'werelogs';
 import Datastore from './Datastore';
 import { generateKey, generateCounter, getCounters, generateStateKey }
@@ -26,8 +25,8 @@ const methods = {
     multiObjectDelete: '_genericPushMetricDeleteObject',
     getObject: '_pushMetricGetObject',
     getObjectAcl: '_genericPushMetric',
-    putObject: '_pushMetricPutObject',
-    copyObject: '_pushMetricCopyObject',
+    putObject: '_genericPushMetricPutObject',
+    copyObject: '_genericPushMetricPutObject',
     putObjectAcl: '_genericPushMetric',
     headBucket: '_genericPushMetric',
     headObject: '_genericPushMetric',
@@ -190,8 +189,8 @@ export default class UtapiClient {
             Object.keys(params).forEach(k => {
                 // Get other properties, but do not include `undefined` ones
                 // or any unrelated properties (those not in `metricProps`).
-                if (props.indexOf(k) < 0 && params[k] !== undefined &&
-                metricProps.indexOf(k) >= 0) {
+                if (props.indexOf(k) < 0 && params[k] !== undefined
+                    && metricProps.indexOf(k) >= 0) {
                     typeObj[k] = params[k];
                 }
             });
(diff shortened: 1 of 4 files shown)


rahulreddy:
@ironman-machine try

src/lib/UtapiClient.js
```diff
@@ -554,163 +623,94 @@ export default class UtapiClient {
     * @param {callback} callback - callback to call
     * @return {undefined}
     */
-    _pushMetricPutObject(params, timestamp, action, log, callback) {
+    _genericPushMetricPutObject(params, timestamp, action, log, callback) {
```
bennettbuchanan: You might include in the JSDoc that this also updates counter for CopyObject action.

src/lib/UtapiClient.js
```diff
@@ -554,163 +623,94 @@ export default class UtapiClient {
...
-                    metric: 'number of objects',
-                    error: actionErr,
-                });
-                return callback(errors.InternalError);
+                ['incrby', generateCounter(p, 'storageUtilizedCounter'),
+                    storageUtilizedDelta],
+                [redisCmd, generateCounter(p, 'numberOfObjectsCounter')],
+                ['incr', generateKey(p, action, timestamp)]
+            );
+            if (action === 'putObject') {
```
bennettbuchanan: nice 👍

src/lib/UtapiClient.js
```diff
@@ -554,163 +623,94 @@ export default class UtapiClient {
...
+                cmds2.push(
+                    ['zremrangebyscore',
+                        generateStateKey(p, 'numberOfObjects'),
+                        timestamp, timestamp],
+                    ['zadd', generateStateKey(p, 'numberOfObjects'),
+                        timestamp, actionCounter]);
+                return true;
+            });
+            if (noErr) {
+                return this.ds.batch(cmds2, callback);
```
bennettbuchanan: Is this not sending metrics in two transactions?

rahulreddy: This step is unavoidable. We have to increment counters, get the results and then use the values to update the sorted sets with the value and timestamp.

bennettbuchanan: ✅

rahulreddy:
@ironman-machine try

rahulreddy:
End to end http://ci.ironmann.io/gh/scality/Integration/9033#build-parameters

src/lib/UtapiClient.js
```diff
@@ -329,40 +328,61 @@ export default class UtapiClient {
...
+                    callback(errors.InternalError);
+                    return false;
+                }
+                cmds2.push(
+                    ['zremrangebyscore', generateStateKey(p, 'storageUtilized'),
+                        timestamp, timestamp],
+                    ['zadd', generateStateKey(p, 'storageUtilized'),
+                        timestamp, actionCounter]
+                );
+                return true;
```
electrachong: if you do this multiple times I feel like you can probably refactor this every check into a util function, e.g. you can pass cmds.length, paramsArr.length, logger, and the types of counters you want to update

rahulreddy: I did think about have a plan for it, just not in this PR. The keys/commands change per method with some nuances, I was waiting for some patterns to emerge before I can extract them into a function

src/lib/UtapiClient.js
```diff
@@ -26,8 +25,8 @@ const methods = {
     multiObjectDelete: '_genericPushMetricDeleteObject',
     getObject: '_pushMetricGetObject',
     getObjectAcl: '_genericPushMetric',
-    putObject: '_pushMetricPutObject',
-    copyObject: '_pushMetricCopyObject',
+    putObject: '_genericPushMetricPutObject',
+    copyObject: '_genericPushMetricPutObject',
```
rahulreddy: We added a generic method for  "putObject" style requests since copyObject differs by only not having to increment `incomingBytes`.

rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine try

rahulreddy:
End to end running here http://ci.ironmann.io/gh/scality/Integration/9180#build-parameters

PR #86, opened 2017-01-24
DEV: Handle error from decodeURIComponent

bennettbuchanan:
Return InvalidURI error if `decodeURIComponent` throws error


bennettbuchanan:
@ironman-machine try

bennettbuchanan:
end-to-end passed

PR #95, opened 2017-01-28
Fwdport forward/rel/6.3 to 6.4

bennettbuchanan:


bennettbuchanan:
e2e: http://ci.ironmann.io/gh/scality/Integration/9042#build-parameters

PR #96, opened 2017-01-30
DEV: Add V4 Auth examples

bennettbuchanan:
Please only review 8a08cb1cb7e4c14955f43edcd8e07c77a6025de2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ const config = {
 }
 const c = new UtapiClient(config);
 
-// The second argument to `PushMetric` is a hexadecimal string Request Unique
+// The second argument to `pushMetric` is a hexadecimal string Request Unique
 // Identifier used for logging.
 c.pushMetric('createBucket', '3d534b1511e5630e68f0', { bucket: 'demo' });
 
@@ -140,11 +140,15 @@ buckets.
 
 There are two options here.
 
-You can generate V4 signature using AWS SDKs. See the following urls for
-reference.
+You can generate V4 signature using AWS SDKs or the node module aws4. See the
+following urls for reference.
 
-http://docs.aws.amazon.com/general/latest/gr/sigv4_signing.html
-http://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html
+* http://docs.aws.amazon.com/general/latest/gr/sigv4_signing.html
+* http://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html
+* https://github.com/mhart/aws4
+
+You may also view examples making a request with Auth V4 using various languages
+and AWS SDKs [here](/examples).
 
 Alternatively, you can use a nifty command line tool available in Scality's S3.
 
--- a/examples/go-V4Auth-example.go
+++ b/examples/go-V4Auth-example.go
@@ -1,0 +1,60 @@
+package main
+
+import (
+    "fmt"
+    "time"
+    "bytes"
+    "encoding/json"
+    "net/http"
+    "io/ioutil"
+    "github.com/aws/aws-sdk-go/aws/credentials"
+    "github.com/aws/aws-sdk-go/aws/signer/v4"
+)
+
+func main() {
+    // Input AWS access key, secret key, and session token.
+    aws_access_key_id := "EO4FRH6BA2L7FCK0EKVT"
+    aws_secret_access_key := "B5QSoChhLVwKzZG1w2fXO0FE2tMg4imAIiV47YWX"
+    token := ""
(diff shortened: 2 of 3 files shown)


examples/go-V4Auth-example.go
```diff
@@ -0,0 +1,51 @@
...
+    enc := json.NewEncoder(buf)
+    enc.Encode(&bucketMetricRequest)
+    request, err := http.NewRequest("POST",
+        fmt.Sprintf("%s/buckets?Action=ListMetrics", "http://localhost:8100"),
+        buf)
+    if err != nil {
+        panic(err)
+    }
+    reader := bytes.NewReader(buf.Bytes())
+    credentials := credentials.NewStaticCredentials("EO4FRH6BA2L7FCK0EKVT",
```
LaurenSpiegel: where are these from?  use a variable so people can insert their own.

examples/go-V4Auth-example.go
```diff
@@ -0,0 +1,51 @@
...
+    "encoding/json"
+    "net/http"
+    "io/ioutil"
+    "github.com/aws/aws-sdk-go/aws/credentials"
+    "github.com/aws/aws-sdk-go/aws/signer/v4"
+)
+
+func main() {
+    bucketName := "test-bucket"
+    d := time.Date(2017, 1, 19, 14, 45, 0, 0, time.UTC)
```
LaurenSpiegel: why starting here?  put comment

examples/go-V4Auth-example.go
```diff
@@ -0,0 +1,51 @@
...
+    credentials := credentials.NewStaticCredentials("EO4FRH6BA2L7FCK0EKVT",
+        "B5QSoChhLVwKzZG1w2fXO0FE2tMg4imAIiV47YWX", "")
+    signer := v4.NewSigner(credentials)
+    signer.Sign(request, reader, "s3", "us-east-1", time.Now())
+    client := &http.Client{}
+    resp, err := client.Do(request)
+    if err != nil {
+        panic(err)
+    }
+    defer resp.Body.Close()
```
LaurenSpiegel: why do you need this?  if this is deferred until the rest of the function returns why not put it at the end so clearer?

examples/go-V4Auth-example.go
bennettbuchanan: This is basically an arbitrary start time. What would be a good time range? (I was thinking perhaps it would be helpful to add an end time as well.)

examples/go-V4Auth-example.go
bennettbuchanan: Indeed, it works just the same if added at the end of the main function. I chose to use `defer` here because that's the pattern used in the docs. (https://golang.org/pkg/net/http/). I could update that though if it's clearer.

examples/go-V4Auth-example.go
LaurenSpiegel: maybe just do 1 month -- all of January or whatever your favorite month is

bennettbuchanan: Agreed! This keeps the response time of the example request constant as this script ages. Also, I have no favorite month, so January it is. 😄

electrachong:
the credentials you use in the example are fake, right? :P

bennettbuchanan:
They are created with vaultclient, they're not AWS credentials.

examples/go-V4Auth-example.go
bennettbuchanan: Thank you, that is more clearn. Updated.

examples/go-V4Auth-example.go
bennettbuchanan: Updated.

README.md
```diff
@@ -44,6 +44,204 @@ c.pushMetric('multiObjectDelete', '3d534b1511e5630e68f0', {
...
+
+You can generate V4 signature using AWS SDKs or the node module aws4. See the
+following urls for reference.
+
+* http://docs.aws.amazon.com/general/latest/gr/sigv4_signing.html
+* http://docs.aws.amazon.com/general/latest/gr/sigv4-signed-request-examples.html
+* https://github.com/mhart/aws4
+
+You may also view examples making a request with Auth V4 using various languages
+and AWS SDKs [here](https://github.com/scality/utapi/tree/DEV/add_example_request_scripts/examples).
```
rahulreddy: You can use a relative url here `/examples`

bennettbuchanan: Updated. Thanks!

rahulreddy:
@bennettbuchanan Go ahead and squash the commits and rebase. I will merge this.

PR #98, opened 2017-02-03
DEV: Allow pushing of unsupported metrics

bennettbuchanan:
Allow passing an unsupported metric to `UtapiClient` to enable merging branches in other repositories first. 

If an unsupported metric is passed in the call to `UtapiClient`, no metrics will be pushed to the Redis datastore.

### DIFF
--- a/src/lib/UtapiClient.js
+++ b/src/lib/UtapiClient.js
@@ -226,14 +226,18 @@ export default class UtapiClient {
     * @return {undefined}
     */
     pushMetric(metric, reqUid, params, cb) {
-        assert(methods[metric], `${metric} metric is not handled by Utapi`);
         const callback = cb || this._noop;
         if (this.disableClient) {
             return callback();
         }
         const log = this.log.newRequestLoggerFromSerializedUids(reqUid);
         const timestamp = UtapiClient.getNormalizedTimestamp();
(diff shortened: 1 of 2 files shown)


src/lib/UtapiClient.js
```diff
@@ -227,14 +227,17 @@ export default class UtapiClient {
...
         if (this.disableClient) {
             return callback();
         }
         const log = this.log.newRequestLoggerFromSerializedUids(reqUid);
         const timestamp = UtapiClient.getNormalizedTimestamp();
-        return this[methods[metric]](params, timestamp, metric, log, callback);
+        if (this[methods[metric]]) {
+            return this[methods[metric]](params, timestamp, metric, log,
+                callback);
+        }
```
electrachong: since we have a logger we might as well log that the metric is not supported, right?

bennettbuchanan: Thanks for the good idea. Updated.

bennettbuchanan:
http://ci.ironmann.io/gh/scality/Integration/9253

rahulreddy:
@bennettbuchanan Please squash and rebase.

bennettbuchanan:
http://ci.ironmann.io/gh/scality/Integration/9329

bennettbuchanan:
e2e passed

PR #108, opened 2017-02-10
FIX: Handle negative counter values

### LINKED ISSUE #107: Handle updates from negative counters on Redis disconnect
Context: While the connection to the Redis server is interrupted, PUT operations are made. Thus `storageUtilized` and `numberOfObjects` metrics are not updated as expected.

The Redis server then reestablishes a connection, and DELETE operations are made. But because metrics were not previously updated for the PUT operations, `storageUtilized` and `numberOfObjects` counters are updated to a negative value. Finally, a PUT operation is made and updates the stateful key values for `storageUtilized` and `numberOfObjects` based on the negative counter values. This causes metrics for the Sorted Set keys to be a negative value.

Duplicate of #105

bennettbuchanan:
Fix #107
* Allow global counter keys to be < 0
* Reset Sorted Set keys as though the bucket
  were empty if the global counter is < 0

### DIFF
--- a/src/lib/UtapiClient.js
+++ b/src/lib/UtapiClient.js
@@ -255,7 +255,9 @@ export default class UtapiClient {
             }
             // storage utilized counter
             const actionErr = results[0][0];
-            const actionCounter = results[0][1];
+            // If < 0, record storageUtilized as though bucket were empty.
+            const actionCounter = results[0][1] < 0 ? objectSize :
+                results[0][1];
             if (actionErr) {
                 log.error('error incrementing counter for push metric', {
                     method: 'UtapiClient.pushMetricUploadPart',
@@ -339,7 +341,8 @@ export default class UtapiClient {
             }
             // number of objects counter
             const actionErr = results[0][0];
-            const actionCounter = results[0][1];
+            // If < 0, record numberOfObjects as though bucket were empty.
+            const actionCounter = results[0][1] < 0 ? 1 : results[0][1];
             if (actionErr) {
                 log.error('error incrementing counter for push metric', {
                     method: 'UtapiClient.pushMetricCompleteMultipartUpload',
@@ -494,6 +497,7 @@ export default class UtapiClient {
             // storage utilized counter
             let actionErr = results[0][0];
             let actionCounter = parseInt(results[0][1], 10);
+            // If < 0, record storageUtilized as though bucket were empty.
             actionCounter = actionCounter < 0 ? 0 : actionCounter;
             if (actionErr) {
                 log.error('error incrementing counter for push metric', {
@@ -514,6 +518,7 @@ export default class UtapiClient {
             // num of objects counter
             actionErr = results[1][0];
             actionCounter = parseInt(results[1][1], 10);
+            // If < 0, record numberOfObjects as though bucket were empty.
             actionCounter = actionCounter < 0 ? 0 : actionCounter;
             if (actionErr) {
                 log.error('error incrementing counter for push metric', {
@@ -684,7 +689,9 @@ export default class UtapiClient {
             let actionCounter;
             // storage utilized counter
             actionErr = results[0][0];
-            actionCounter = results[0][1];
+            // If < 0, record storageUtilized as though bucket were empty.
+            actionCounter = results[0][1] < 0 ? storageUtilizedDelta :
+                results[0][1];
             if (actionErr) {
                 log.error('error incrementing counter for push metric', {
                     method: 'UtapiClient.pushMetricPutObject',
@@ -702,7 +709,8 @@ export default class UtapiClient {
 
             // number of objects counter
             actionErr = results[1][0];
-            actionCounter = results[1][1];
+            // If < 0, record numberOfObjects as though bucket were empty.
+            actionCounter = results[1][1] < 0 ? 1 : results[1][1];
             if (actionErr) {
                 log.error('error incrementing counter for push metric', {
                     method: 'UtapiClient.pushMetricPutObject',
@@ -771,10 +779,11 @@ export default class UtapiClient {
             }
             const cmds = [];
             let actionErr;
-            let actionCounter;
             // storage utilized counter
             actionErr = results[0][0];
-            actionCounter = results[0][1];
+            // If < 0, record storageUtilized as though bucket were empty.
+            let actionCounter = results[0][1] < 0 ? storageUtilizedDelta :
+                results[0][1];
             if (actionErr) {
                 log.error('error incrementing counter for push metric', {
                     method: 'UtapiClient.pushMetricCopyObject',
@@ -792,7 +801,8 @@ export default class UtapiClient {
 
             // number of objects counter
             actionErr = results[1][0];
-            actionCounter = results[1][1];
+            // If < 0, record numberOfObjects as though bucket were empty.
+            actionCounter = results[1][1] < 0 ? 1 : results[1][1];
             if (actionErr) {
                 log.error('error incrementing counter for push metric', {
                     method: 'UtapiClient.pushMetricCopyObject',
--- a/src/lib/schema.js
+++ b/src/lib/schema.js
@@ -74,12 +74,13 @@ export function getBucketKeys(bucket, timestamp) {
 /**
 * Returns metric from key
 * @param {string} key - schema key
-* @param {string} bucket - bucket name
-* @return {string} metric - S3 metric
+* @return {string} metric - Utapi metric
 */
-export function getMetricFromKey(key, bucket) {
-    // s3:buckets:1473451689898:demo:putObject
-    return key.slice(25).replace(`${bucket}:`, '');
+export function getMetricFromKey(key) {
+    const fields = key.split(':');
+    // Identify the location of the metric in the array.
+    const metricLocation = key.includes('counter') ? -2 : -1;
+    return fields[fields.length + metricLocation];
 }
 
 /**
--- a/tests/functional/testUtapiClient.js
+++ b/tests/functional/testUtapiClient.js
@@ -1,37 +1,89 @@
 import assert from 'assert';
-import { mapSeries, series } from 'async';
+import { map, series } from 'async';
 import UtapiClient from '../../src/lib/UtapiClient';
-import MemoryBackend from '../../src/lib/backend/Memory';
 import Datastore from '../../src/lib/Datastore';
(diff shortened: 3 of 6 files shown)


tests/functional/testUtapiClient.js
```diff
@@ -1,37 +1,94 @@
...
+        datastore.get(item, (err, res) => {
             if (err) {
                 return next(err);
             }
-            const metric = getMetricFromKey(item, bucket)
-                .replace(':counter', '');
-            assert.equal(res, 0, `${metric} must be 0`);
+            let metric = getMetricFromStateKey(item, bucket);
+            metric = metric.replace(':counter', '');
+            assert.equal(res, valueObj[metric],
```
rahulreddy: Please change this to use `assert.strictEqual` here. Your tests are actually comparing a string to a number.
Pasting a console.log here
```
res [ '9' ] valueObj[metric] 9
res [ '1' ] valueObj[metric] 1
```

tests/functional/testUtapiClient.js
```diff
@@ -1,37 +1,94 @@
...
+                `${metric} must be ${valueObj[metric]}`);
             return next();
         }), cb);
 }
 
+function _assertStateKeys(bucket, valueObj, cb) {
+    const stateKeys = getBucketStateKeys(testBucket);
+    return mapSeries(stateKeys, (item, next) =>
```
rahulreddy: Why not just use the map, it would execute the queries in parallel.

src/lib/UtapiClient.js
```diff
@@ -255,7 +255,8 @@ export default class UtapiClient {
             }
             // storage utilized counter
             const actionErr = results[0][0];
-            const actionCounter = results[0][1];
+            const actionCounter = results[0][1] < 0 ? objectSize :
```
rahulreddy: Please add a one-liner comment explaining why objectSize was used vs 0.

tests/functional/testUtapiClient.js
```diff
@@ -49,6 +106,53 @@ describe('Counters', () => {
...
             next => utapiClient.pushMetricCreateBucket(reqUid, testBucket,
                 next),
-        ], () => _assertCounters(testBucket, done));
+        ], () => _assertCounters(testBucket, {
+            storageUtilized: 0,
+            numberOfObjects: 0,
+        }, done));
     });
+
+    it('should update counters to correct values < 0: putObject', done =>
```
rahulreddy: Since we are technically not updating counters but actually recording the counter for a timestamp, the test description should reflect that.

tests/functional/testUtapiClient.js
```diff
@@ -1,37 +1,94 @@
...
+        next => {
+            if (operation === 'PutObject') {
+                return utapiClient.pushMetricPutObject(reqUid, testBucket, 9,
+                    null, next);
+            }
+            if (operation === 'CopyObject') {
+                return utapiClient.pushMetricCopyObject(reqUid, testBucket, 9,
+                    null, next);
+            }
+            return utapiClient.pushMetricUploadPart(reqUid, testBucket, 9,
```
jonathan-gramain: Could be clearer to check the operation type here too (with a switch?)

tests/functional/testUtapiClient.js
```diff
@@ -49,6 +106,53 @@ describe('Counters', () => {
...
+        _checkMissingOperations(_assertCounters, 'CopyObject', counterKeyValues,
+            done));
+
+    it('should update counters to correct values < 0: completeMPU', done =>
+        _checkMissingOperations(_assertCounters, 'CompleteMultipartUpload',
+            counterKeyValues, done));
+
+    it('should update counters to correct values < 0: uploadPart', done =>
+        // Uploading a part does not increment numberOfObjects value.
+        _checkMissingOperations(_assertCounters, null,
```
jonathan-gramain: related to my previous comment: I think passing the operation name here would be more readable.

bennettbuchanan: Thank you. I had an internal debate about this, so I will update as you suggest. :)

src/lib/schema.js
```diff
@@ -83,6 +83,17 @@ export function getMetricFromKey(key, bucket) {
...
 
 /**
+* Returns metric from a state key
+* @param {string} key - schema key
+* @param {string} bucket - bucket name
+* @return {string} metric - S3 metric
+*/
+export function getMetricFromStateKey(key, bucket) {
+    // s3:buckets:foo:storageUtilized
+    return key.slice(11).replace(`${bucket}:`, '');
```
jonathan-gramain: Could return what is after the last colon, otherwise if that does not work, you may use `'s3:buckets'.length` instead of the hard-coded 11 after checking that's how the string starts (unless there's a potential big performance impact doing this).

bennettbuchanan: Updated. That's much more intuitive—I will eventually update those other methods using the same pattern. Thanks!

tests/functional/testUtapiClient.js
bennettbuchanan: Updated.

tests/functional/testUtapiClient.js
bennettbuchanan: Updated.

src/lib/UtapiClient.js
bennettbuchanan: I've done this for each pushMetric method.

tests/functional/testUtapiClient.js
bennettbuchanan: Updated.

tests/functional/testUtapiClient.js
bennettbuchanan: Now checking for integer values in both assertions.

bennettbuchanan:
http://ci.ironmann.io/gh/scality/Integration/9502

tests/functional/testUtapiClient.js
bennettbuchanan: Updated.

src/lib/schema.js
```diff
@@ -83,6 +83,17 @@ export function getMetricFromKey(key, bucket) {
...
 
 /**
+* Returns metric from a state key
+* @param {string} key - schema key
+* @param {string} bucket - bucket name
+* @return {string} metric - S3 metric
+*/
+export function getMetricFromStateKey(key, bucket) {
+    // s3:buckets:foo:storageUtilized
+    return key.slice('s3:buckets:'.length).replace(`${bucket}:`, '');
```
rahulreddy: One last change, this can be a constant up top as it's static.

bennettbuchanan: As in, no need for the method?

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

src/lib/schema.js
```diff
@@ -39,6 +39,12 @@ const bucketKeys = {
     outgoingBytes: (b, t) => `s3:buckets:${t}:${b}:outgoingBytes`,
 };
 
+const metric = {
+    stateKey: (k, b) => k.slice('s3:buckets:'.length).replace(`${b}:`, ''),
+    timestampKey: (k, b) =>
+        k.slice('s3:buckets:0000000000000:'.length).replace(`${b}:`, ''),
```
jonathan-gramain: I fear this is a bit fragile to rely on the number of digits in a timestamp. Since fields look to be delimited by colons, and since you must know how many fields you want to truncate, can you instead slice up to the number of fields you want to remove, where they are separated by colons? You can consider using a regexp if that makes the thing simpler.

tests/functional/testUtapiClient.js
```diff
@@ -1,37 +1,90 @@
...
+
+datastore.setClient(redis);
 utapiClient.setDataStore(datastore);
 
-function _assertCounters(bucket, cb) {
+function _assertCounters(bucket, valueObj, cb) {
     const counters = getBucketCounters(testBucket);
-    return mapSeries(counters, (item, next) =>
-        memBackend.get(item, (err, res) => {
+    return map(counters, (item, next) =>
```
jonathan-gramain: minor: `next` could be called simply `cb` or `done`, because calls are neither ordered nor serialized (anymore).

tests/functional/testUtapiClient.js
```diff
@@ -1,37 +1,90 @@
...
+            metric = metric.replace(':counter', '');
+            assert.strictEqual(parseInt(res, 10), valueObj[metric],
+                `${metric} must be ${valueObj[metric]}`);
+            return next();
+        }), cb);
+}
+
+function _assertStateKeys(bucket, valueObj, cb) {
+    const stateKeys = getBucketStateKeys(testBucket);
+    return map(stateKeys, (item, next) =>
```
jonathan-gramain: same here

src/lib/UtapiClient.js
```diff
@@ -339,7 +341,8 @@ export default class UtapiClient {
             }
             // number of objects counter
             const actionErr = results[0][0];
-            const actionCounter = results[0][1];
+            // If < 0, record numberOfObjects as though bucket were empty.
```
jonathan-gramain: I have to admit the logic of handling negative results escapes me, but I trust it's the right thing :)

src/lib/schema.js
rahulreddy: I agree, slicing it up is more predictable.

bennettbuchanan:
@ironman-machine try

src/lib/UtapiClient.js
bennettbuchanan: If UtapiReplay does not update previously unsuccessful metrics when a 15 minute interval occurs, there is the possibility of the metrics showing a negative value for that interval. Of course, it doesn't ever make sense for `storageUtilized` or `numberOfObjects` to be recorded as a negative value.

This addresses that possibility by instead recording a value of `0` for that time period, essentially allowing some flexibility. The design is such that eventually these metrics will be updated appropriately.

tests/functional/testUtapiClient.js
bennettbuchanan: Updated.

tests/functional/testUtapiClient.js
bennettbuchanan: Updated.

src/lib/schema.js
bennettbuchanan: Thanks, that's much better. 😄

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
e2e passed

PR #109, opened 2017-02-15
FT: Add Utapi replay

bennettbuchanan:
Update `UtapiClient` to push metrics to a local cache (Redis List datatype) if an internal error occurs (for example, in a case where we are unable to connect to the Redis server).

Add a class `UtapiReplay` that attempts to push any cached metrics with `UtapiClient` at the given interval or schedule.

Depends on https://github.com/scality/S3/pull/612

PR #110, opened 2017-02-15
FT: Add UtapiReplay to DESIGN.md and README.md

bennettbuchanan:
Dependent on https://github.com/scality/utapi/pull/109.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -40,9 +40,10 @@ between 06:15:00 and 06:29:59.
 ### Data Storage
 
 Utapi uses Redis as a database for storage of its metrics values. Accordingly,
-it uses two different Redis data types: Sorted Sets and Strings. This document
-describes how these two data types are used by Utapi. For further information on
-data types see the Redis [documentation](https://redis.io/topics/data-types).
+it uses three different Redis data types: Sorted Sets, Strings, and Lists. This
+document describes how these three data types are used by Utapi. For further
+information on data types see the Redis
+[documentation](https://redis.io/topics/data-types).
 
 #### Sorted Sets
 
@@ -59,9 +60,9 @@ values.
 
 #### Strings
 
-The remaining Redis keys recorded by Utapi use a String data type. These include
-metrics for incoming bytes, outgoing bytes, and all S3 operations (e.g.,
-'CreateBucket', 'PutObject', etc.).
+The Redis keys storing metrics for incoming bytes, outgoing bytes, and all S3
+operations (e.g., 'CreateBucket', 'PutObject', etc.) are recorded with a String
+data type.
 
 Moreover, there are also global counters associated with metrics for the number
 of objects and the storage utilized. Such counters are updated during any
@@ -70,9 +71,16 @@ counter for the number of objects increments or decrements, respectively. These
 counters are used internally by Sorted Sets to record the state (the storage
 used and the number of objects) at a particular point in time.
 
+#### Lists
+
+Redis keys storing cached metrics use a List data type, where each List element
+is a string containing information from the original request. This datatype is
+used by a component named UtapiReplay that pushes any metrics stored in the List
+every five minutes, by default.
+
 #### Example
 
-Steps occurring during a 'PutObject' request:
+Steps occurring during a successful 'PutObject' request:
 
 1. If the new object overwrites a pre-existing object, the counter for the
     number of objects remains unchanged. Otherwise it increments by one.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -50,6 +50,12 @@ c.pushMetric('multiObjectDelete', '3d534b1511e5630e68f0', {
 });
 ```
 
+If an error occurs during and the client is unable to record metrics in the
+underlying datastore, metric data is instead stored in a local cache. Utapi
+attempts to push these cached metrics every five minutes, by default. The
+particularities of this behavior is configurable. For further information, see
```
dora-korpar: the particularities.... /are/ configurable

bennettbuchanan: Updated.

DESIGN.md
```diff
@@ -127,6 +136,16 @@ the operation):
...
 ```
 
+#### Local Cache Key
+
+Metrics of operations pushed by S3 that are unsuccessfully recorded as schema
+keys in the Redis datastore (for example, in the case of a failed connection)
+are stored in a key:
+
+```
+s3:utapireplay
```
LaurenSpiegel: what does an example value look like?

README.md
```diff
@@ -50,6 +50,12 @@ c.pushMetric('multiObjectDelete', '3d534b1511e5630e68f0', {
 });
 ```
 
+If an error occurs during and the client is unable to record metrics in the
```
LaurenSpiegel: during what?

README.md
```diff
@@ -50,6 +50,12 @@ c.pushMetric('multiObjectDelete', '3d534b1511e5630e68f0', {
...
+underlying datastore, metric data is instead stored in a local cache. Utapi
```
LaurenSpiegel: local redis cache?

README.md
```diff
@@ -50,6 +50,12 @@ c.pushMetric('multiObjectDelete', '3d534b1511e5630e68f0', {
...
+attempts to push these cached metrics every five minutes, by default. The
```
LaurenSpiegel: what happens if the push at a 5 minute interval fails and continues to fail?  Is there any alert action taken?

DESIGN.md
```diff
@@ -70,9 +71,14 @@ counter for the number of objects increments or decrements, respectively. These
 counters are used internally by Sorted Sets to record the state (the storage
 used and the number of objects) at a particular point in time.
 
+#### Lists
+
+Redis keys storing cached metrics use a List data type, where each List element
```
LaurenSpiegel: you should mention the utapiReplay component here

DESIGN.md
```diff
@@ -90,6 +96,9 @@ Steps occurring during a 'PutObject' request:
 5. The Sorted Set keys (the storage used and the number of objects) are updated
     to the value of their respective counter.
 
+If a connection to the Redis datastore cannot be made, metrics from the original
+request to Utapi are pushed to a local cache to be retried at a later time.
```
LaurenSpiegel: a local redis cache.

DESIGN.md
bennettbuchanan: Updated.

DESIGN.md
bennettbuchanan: I've added an example of a List element.

README.md
bennettbuchanan: Nice catch! Updated.

README.md
bennettbuchanan: Updated.

DESIGN.md
bennettbuchanan: Updated.

DESIGN.md
```diff
@@ -127,6 +139,24 @@ the operation):
 s3:buckets:1451635200000:foo-bucket:PutObject
 ```
 
+#### Local Redis Cache Key
+
```
rahulreddy: Can you expand the definition so that it's clear that this key is created in local cache and not Redis datastore.

README.md
bennettbuchanan: I've added an explanation here. However, there is no special alert action for such a case, aside from the regular standard behavior of UtapiClient logging an `InternalError` and `UtapiReplay` logging the actions.

DESIGN.md
bennettbuchanan: Good point.

bennettbuchanan: Updated.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

Issue #39, opened 2016-09-10
Abort Multipartupload should decrease the storage utilization


rahulreddy:
pushMetricAbortMultipartUpload action in UtapiClient has to take storage reduced as a param

PR #111, opened 2017-02-17
FIX: Update storageUtilized on abortMPU

bennettbuchanan:
Require a `byteLength` property with an abortMPU operation to enable correct updating of the `storageUtilized` value.

Fix https://github.com/scality/utapi/issues/39.

### DIFF
--- a/src/lib/UtapiClient.js
+++ b/src/lib/UtapiClient.js
@@ -23,7 +23,7 @@ const methods = {
     completeMultipartUpload: '_pushMetricCompleteMultipartUpload',
     listMultipartUploads: '_pushMetricListBucketMultipartUploads',
     listMultipartUploadParts: '_genericPushMetric',
-    abortMultipartUpload: '_genericPushMetric',
+    abortMultipartUpload: '_genericPushMetricDeleteObject',
     deleteObject: '_genericPushMetricDeleteObject',
     multiObjectDelete: '_genericPushMetricDeleteObject',
     getObject: '_pushMetricGetObject',
@@ -488,7 +488,8 @@ export default class UtapiClient {
     }
 
     /**
-    * Updates counter for DeleteObject or MultiObjectDelete action
+    * Updates counter for DeleteObject, MultiObjectDelete, or
+    * AbortMultipartUpload action
     * @param {object} params - params for the metrics
     * @param {string} [params.bucket] - (optional) bucket name
     * @param {string} [params.accountId] - (optional) account ID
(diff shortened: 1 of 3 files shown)


rahulreddy:
@bennettbuchanan Please re-target this to 6.3

src/lib/UtapiClient.js
```diff
@@ -478,6 +478,72 @@ export default class UtapiClient {
...
+        ));
+        const cmdsLen = cmds.length;
+        const paramsArrLen = paramsArr.length;
+        return this.ds.batch(cmds, (err, results) => {
+            if (err) {
+                log.error('error incrementing counter', {
+                    method: 'UtapiClient._pushMetricAbortMultipartUpload',
+                    error: err,
+                });
+                return callback(errors.InternalError);
```
LaurenSpiegel: why not returning false like below?  why are you returning booleans from this function? do you use them?  the jsdoc says return is undefined and it is not consistent what is being returned.

tests/unit/testUtapiClient.js
```diff
@@ -240,9 +240,16 @@ describe('UtapiClient:: push metrics', () => {
...
-        const expected = getObject(timestamp,
-            { action: 'AbortMultipartUpload' });
-        testMetric('abortMultipartUpload', metricTypes, expected, done);
+        const expected = getObject(timestamp, {
+            action: 'AbortMultipartUpload',
+            storageUtilized: '0',
+        });
+        Object.assign(params, metricTypes, { byteLength: 1024 });
+        // Set mock data of one, 1024 byte part object for
+        // `AbortMultipartUpload` to update.
```
LaurenSpiegel: so you are setting the storageUtilized at 1024 and then subtracting it with the params argument with abortMultipartUpload? Just clarifying.

rahulreddy:
Ideally this should have been `6.3` but since the client changed a lot from 6.3 to 6.4, we will add this fix to 6.4 and backport to 6.3 if needed.

tests/unit/testUtapiClient.js
bennettbuchanan: Exactly! 😄

bennettbuchanan: Exactly! 😄

src/lib/UtapiClient.js
bennettbuchanan: I've updated this to reuse a pre-existing method `_genericPushMetricDeleteObject` to cut back on code duplication. However, this method and many others in UtapiClient also have that same pattern in the JSDOC. Perhaps these should all be updated together?

src/lib/UtapiClient.js
```diff
@@ -555,16 +559,19 @@ export default class UtapiClient {
...
-                objectsIndex = i * (cmdsLen / paramsArrLen) + 1;
+                objectsIndex = i * (cmdsLen / paramsArrLen) + 2;
                 actionErr = results[objectsIndex][0];
                 actionCounter = parseInt(results[objectsIndex][1], 10);
                 actionCounter = actionCounter < 0 ? 0 : actionCounter;
                 if (actionErr) {
                     log.error('error incrementing counter for push metric', {
-                        method: 'UtapiClient._genericPushMetricDeleteObject',
-                        metric: 'num of objects',
+                        method: 'UtapiClient._genericPushMetricDelete Object',
```
rahulreddy: remove space in log`_genericPushMetricDelete Object`

src/lib/UtapiClient.js
```diff
@@ -555,16 +559,19 @@ export default class UtapiClient {
...
                         generateStateKey(p, 'storageUtilized'),
                         timestamp, actionCounter]);
-
+                // Abort multipart upload does not affect number of objects
+                if (action === 'abortMultipartUpload') {
+                    return true;
+                }
                 // num of objects counter
-                objectsIndex = i * (cmdsLen / paramsArrLen) + 1;
+                objectsIndex = i * (cmdsLen / paramsArrLen) + 2;
```
LaurenSpiegel: As discussed, either add a comment or define objectIncrementer as cmds[2] or something like that so we know what the magical +2 refers to.

bennettbuchanan: I've attempted to clarify these values.

bennettbuchanan:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=dev/check_storageUtilized_abort_MPU
try

bennettbuchanan:
e2e passed with associated Integration tests. http://ci.ironmann.io/gh/scality/Integration/9847

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

PR #112, opened 2017-02-17
ft: support out of order operations

rahulreddy:
Former logic was setting counters to 0 on bucket creation which would
affect counters when push metric requests come in out of order,
for example, delete object AA -> put object AA -> create bucket A. This would
happen if Redis Sentinel is down, the metrics are cached temporarily and replayed
later.


rahulreddy:
@ironman-machine try

bennettbuchanan:
Passed here: http://ci.ironmann.io/gh/scality/Integration/9657#build-parameters

Issue #52, opened 2016-09-29
Add a route for getting the latest metrics


rahulreddy:
Add a new route which will provide metrics based on the ongoing activity. The route would not take any time range as input and use current time as start time.

bennettbuchanan:
Added by https://github.com/scality/utapi/pull/114.

PR #114, opened 2017-02-24
FT: Add recent metrics route

bennettbuchanan:
Address #52.

### DIFF
--- a/src/handlers/metricsHandlers.js
+++ b/src/handlers/metricsHandlers.js
@@ -1,5 +1,14 @@
 import ListMetrics from '../lib/ListMetrics';
 
+function listRecentMetrics(utapiRequest, metric, service, cb) {
+    const log = utapiRequest.getLog();
+    log.debug('handling list recent metrics request', {
+        method: 'listRecentMetrics',
+    });
+    const Metric = new ListMetrics(metric, service);
+    return Metric.getRecentTypesMetrics(utapiRequest, cb);
+}
+
 /**
  * @class BucketsHandler
  * Handles Buckets resource actions
@@ -20,6 +29,18 @@ export class BucketsHandler {
         });
         const Buckets = new ListMetrics('buckets', service);
         return Buckets.getTypesMetrics(utapiRequest, cb);
+    }
+
+    /**
+    * List metrics starting from the second most recent fifteen minute timestamp
+    * for the given list of buckets
+    * @param {UtapiRequest} utapiRequest - UtapiRequest instance
+    * @param {string} service - the service component (e.g., 's3')
+    * @param {callback} cb - callback
+    * @return {undefined}
+    */
+    static listRecentMetrics(utapiRequest, service, cb) {
+        return listRecentMetrics(utapiRequest, 'buckets', service, cb);
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/handlers/metricsHandlers.js
```diff
@@ -20,6 +29,16 @@ export class BucketsHandler {
...
         return Buckets.getTypesMetrics(utapiRequest, cb);
     }
+
+    /**
+    * List metrics for the most recent 15 minutes of a given list of buckets
+    * @param {UtapiRequest} utapiRequest - UtapiRequest instance
+    * @param {callback} cb - callback
+    * @return {undefined}
+    */
+    static listRecentMetrics(utapiRequest, cb) {
```
rahulreddy: These handlers duplicate a lot of code. Can you see if you can use one class with a static method and parameterize the metric level (service/accounts/buckets/users)

bennettbuchanan:
@rahulreddy I am wondering now if it would be helpful to instead list metrics beginning from the previous fifteen minute timestamp (i.e., if it's 6:35 when the request is made, then list metrics starting from 6:15). This would avoid a case where the listing of recent metrics would _seemingly_ not show the most recent metrics. For example, a user puts an object at 6:29, and lists recent metrics at 6:30.

bennettbuchanan:
@ironman-machine

bennettbuchanan:
@ironman-machine

src/validators/listMetrics.js
```diff
@@ -4,11 +4,13 @@ import Validator from './Validator';
...
  * @param {object} dict - Input fields for route
+ * @param {boolean} [isRecentListing] - `true` if the route is for listing
+ * recent metrics, otherwise `undefined`
  * @return {Validator} Return the created validator
  */
-function getValidator(metricType, dict) {
-    const obj = {};
-    obj.timeRange = true;
+function getValidator(metricType, dict, isRecentListing) {
+    // Do not validate time ranges for recent listings because Utapi creates it
```
electrachong: Just checking the comment was finished? Utapi creates the time ranges you mean?

src/validators/listMetrics.js
```diff
@@ -4,11 +4,13 @@ import Validator from './Validator';
  * Get the validator for a particular metric type
  * @param {string} metricType - metric type (e.g., 'buckets', 'accounts')
  * @param {object} dict - Input fields for route
+ * @param {boolean} [isRecentListing] - `true` if the route is for listing
+ * recent metrics, otherwise `undefined`
```
electrachong: By recent listings, do you want to specify within the most recent 15 minutes?

src/handlers/metricsHandlers.js
```diff
@@ -67,6 +98,17 @@ export class ServiceHandler {
         const Accounts = new ListMetrics('service', service);
         return Accounts.getTypesMetrics(utapiRequest, cb);
     }
+
+    /**
+    * List metrics for the most recent 15 minutes of a given list of accounts
```
electrachong: given service?

bennettbuchanan:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=ft/S3C-78-add-recent-metrics-route-tests
try

bennettbuchanan:
I've increased the range to start from the second most recent normalized timestamp (i.e., the definition of "recent" is now a time range up to 29:59:999).

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
e2e passed

PR #117, opened 2017-03-09
FT: Add service-level metrics

bennettbuchanan:
Add service-level metrics support, the service-level is set using a property `component` passed to the Utapi configuration (see scality/S3#537).


bennettbuchanan:
@ironman-machine 
SCALITY_INTEGRATION_BRANCH=test/add_service_level_utapi_tests
try

src/lib/ListMetrics.js
```diff
@@ -200,13 +205,14 @@ export default class ListMetrics {
                         cmd: key,
                     });
                 } else {
-                    const m = getMetricFromKey(key, resource, this.metric);
+                    const m = getMetricFromKey(key);
```
bennettbuchanan: This method was updated in a previous PR, hence remove unnecessary parameters.

src/lib/UtapiClient.js
```diff
@@ -435,14 +451,14 @@ export default class UtapiClient {
...
             let actionErr;
             let actionCounter;
             let key;
-            const paramsArrLen = paramsArr.length;
             const cmds2 = [];
             const noErr = paramsArr.every((p, i) => {
-                index = i * paramsArrLen;
+                // We want the first element of every group of two commands
+                // returned from ioredis.
+                const index = i * commandsGroupSize;
```
bennettbuchanan: Update to commandsGroupSize since if paramsArrLen is greater than 2 (which is the case when adding the service level), we will be off by one.

src/lib/UtapiClient.js
```diff
@@ -85,6 +85,11 @@ export default class UtapiClient {
                 this.localCache = new Datastore()
                     .setClient(redisClient(config.localCache, this.log));
             }
+            if (config.component) {
```
rahulreddy: Please add this to the JSDoc above.

src/lib/UtapiClient.js
```diff
@@ -227,26 +240,26 @@ export default class UtapiClient {
...
-        }
-        if (params.accountId && this.metrics.indexOf('accounts') >= 0) {
-            props.push(Object.assign({
-                accountId: params.accountId,
-                level: 'accounts',
-            }, metricData));
-        }
+        // We add a `service` property to any non-service level to be able to
+        // build the appropriate schema key.
+        this.metrics.forEach(level => {
```
rahulreddy: Please use map here and return the obj directly.

src/lib/UtapiClient.js
```diff
@@ -424,6 +437,9 @@ export default class UtapiClient {
                 ['incr', generateKey(p, action, timestamp)]
             );
         });
+        // We track the number of commands needed for each `paramsArr` object to
+        // eventually locate each group in the results from ioredis.
```
rahulreddy: results from ~~ioredis~~ redis

src/lib/UtapiClient.js
```diff
@@ -435,14 +451,14 @@ export default class UtapiClient {
...
-            let index;
             let actionErr;
             let actionCounter;
             let key;
-            const paramsArrLen = paramsArr.length;
             const cmds2 = [];
             const noErr = paramsArr.every((p, i) => {
-                index = i * paramsArrLen;
+                // We want the first element of every group of two commands
+                // returned from ioredis.
```
rahulreddy: returned from ~~ioredis~~ redis.

src/lib/UtapiClient.js
```diff
@@ -424,6 +437,9 @@ export default class UtapiClient {
                 ['incr', generateKey(p, action, timestamp)]
             );
         });
+        // We track the number of commands needed for each `paramsArr` object to
+        // eventually locate each group in the results from ioredis.
+        const commandsGroupSize = 2;
```
rahulreddy: Shouldn't this be `paramsArr.length` ?

tests/unit/testMetrics.js
```diff
@@ -4,28 +4,31 @@ import Datastore from '../../src/lib/Datastore';
...
 // Create the metric response object for a given metric.
 function getMetricResponse(schemaKey) {
     // Use `JSON.parse` to make deep clone because `Object.assign` will
     // copy property values.
-    const response = JSON.parse(JSON.stringify(metricResponseJSON));
+    const response = JSON.parse(JSON.stringify(s3metricResponseJSON));
     const responseKeys = {
         bucket: 'bucketName',
         accountId: 'accountId',
+        service: 'serviceName',
```
rahulreddy: This changes the defined spec. I updated the specification docs. to reflect the same.

tests/unit/testUtapiClient.js
```diff
@@ -17,28 +17,25 @@ const redisLocal = { host: 'localhost', port: 6379 };
...
+        {
+            key: 's3:buckets:foo-bucket',
+            timestampKey: `s3:buckets:${timestamp}:foo-bucket`,
+        },
+        {
+            key: 's3:accounts:foo-account',
+            timestampKey: `s3:accounts:${timestamp}:foo-account`,
+        },
+        {
+            key: 's3:service:s3',
```
rahulreddy: Why not just `s3:service`. If any other service ends up using Utapi, it will be namespaced accordingly, for example `metadata:service`.

src/lib/UtapiClient.js
bennettbuchanan: It should be the length of the group of commands pushed to `cmds` for each object in `paramsArr`. There is a problem when locating the values in the response from Redis if `paramsArr.length` is > 2, at least in calculate their location using the `every` method here: https://github.com/scality/utapi/pull/117/files/26f297d3d7f986f1c1eec96da387bcd8551abee7#diff-ca6d1c871f1efa6d1be088e4ec103337L444

src/lib/UtapiClient.js
bennettbuchanan: `forEach` was used since there could be a case in which a metric level is not included in the `params`, but is included in `this.metrics`. Such a case would result in an array returned with an `undefined` element.

src/lib/UtapiClient.js
bennettbuchanan: Updated. Also added a default component of `'s3'`.

src/lib/UtapiClient.js
bennettbuchanan: Updated.

src/lib/UtapiClient.js
bennettbuchanan: Updated.

bennettbuchanan:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=test/add_service_level_utapi_tests
try

bennettbuchanan:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=test/add_service_level_utapi_tests
try after squash

src/handlers/metricsHandlers.js
```diff
@@ -9,15 +9,16 @@ export class BucketsHandler {
     /**
     * List metrics for the given list of buckets
     * @param {UtapiRequest} utapiRequest - UtapiRequest instance
+    * @param {string} service - the service component (e.g., 's3')
```
dora-korpar: What are other services this is/will be needed for?

src/lib/UtapiClient.js
```diff
@@ -227,26 +237,26 @@ export default class UtapiClient {
...
-                accountId: params.accountId,
-                level: 'accounts',
-            }, metricData));
-        }
+        // We add a `service` property to any non-service level to be able to
+        // build the appropriate schema key.
+        this.metrics.forEach(level => {
+            const prop = metricObj[level];
+            // There will be no `service` property of the params.
+            if (params[prop] || level === 'service') {
```
dora-korpar: Nice bit of logic here :)

src/handlers/metricsHandlers.js
bennettbuchanan: Yes, could be. This allows for an endless amount of creativity! 😄

bennettbuchanan:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=test/add_service_level_utapi_tests
try

rahulreddy:
@ironman-machine
SCALITY_INTEGRATION_BRANCH=test/add_service_level_utapi_tests
try

PR #121, opened 2017-03-17
fix: remove hiredis module

rahulreddy:
This module was intended for faster parsing of heavy responses from Redis.
In our use case, the default parser is fast enough and ioredis is ditching this
module in favor of the default parser in the next version.


rahulreddy:
@ironman-machine try

PR #123, opened 2017-03-31
fix: pad insufficient data points

rahulreddy:
This fix adds padding for insufficient data points i.e., when storageUtilized
or numberOfObject sorted sets have only one element.


rahulreddy:
@ironman-machine try

src/lib/ListMetrics.js
```diff
@@ -191,8 +196,26 @@ export default class ListMetrics {
             const metricResponse = this._getMetricResponse(resource, start,
                 end);
             // last 4 are results of storageUtilized, numberOfObjects,
```
bennettbuchanan: Comment needs updating for new storageUtilizedCount and numberOfObjectsCount values.

bennettbuchanan (approved):
Great catch! Minor comment.

rahulreddy:
This solution is insufficient. I will revisit this once the need arises.

PR #129, opened 2017-05-06
FT: Handle metrics for versioning

bennettbuchanan:
* Add support for `PutBucketVersioning ` and  `GetBucketVersioning` metrics.
* Add a method `_pushMetricDeleteMarkerObject ` which handles metrics for putting a delete marker when a bucket has versioning enabled.


bennettbuchanan:
@ironman-machine try

src/lib/UtapiClient.js
```diff
@@ -335,6 +338,74 @@ export default class UtapiClient {
...
+                // We want the first element of every group of two commands
+                // returned from Redis. This contains the value of the
+                // numberOfObjectsCounter after it has been incremented.
+                const index = i * commandsGroupSize;
+                actionErr = results[index][0];
+                actionCounter = parseInt(results[index][1], 10);
+                // If < 0 or NaN, record numberOfObjects as though bucket were
+                // empty.
+                actionCounter = Number.isNaN(actionCounter) ||
+                    actionCounter < 0 ? 1 : actionCounter;
```
electrachong: so if the bucket is empty, you increment by one?

src/lib/UtapiClient.js
```diff
@@ -335,6 +338,74 @@ export default class UtapiClient {
...
+                if (actionErr) {
```
electrachong: you can bump this ahead of the actionCounter part since you return in case of error, right?

tests/unit/testUtapiClient.js
```diff
@@ -408,6 +408,22 @@ describe('UtapiClient:: push metrics', () => {
         testMetric('headObject', metricTypes, expected, done);
     });
 
+    it('should push putBucketVersioning metrics', done => {
```
electrachong: no need for a test to push get bucket versioning metrics?

bennettbuchanan: Good catch.

src/lib/UtapiClient.js
bennettbuchanan: Thank you. I've updated this.

src/lib/UtapiClient.js
bennettbuchanan: I noticed in S3 we put a delete marker whether a marker exists or not. Is this correct? If so, there could be a case where the bucket is empty yet a delete marker is put.

bennettbuchanan:
@ironman-machine try

bennettbuchanan:
@ironman-machine try

PR #133, opened 2017-05-23
chore: remove babel

rahulreddy:
- Change ES6 imports to require
- Remove post-install step which compiles the code using babel
- Remove babel modules
- Remove src, dist directories as we no longer compile


rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine try

PR #135, opened 2017-05-24
FT: Add example of using Account ID in ARN

bennettbuchanan:
Depends on:
* https://github.com/scality/Arsenal/pull/260
* https://github.com/scality/Integration/pull/530

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ Server is running.
     aws iam --endpoint-url <endpoint> create-access-key --user-name utapiuser
     ```
 
-3. Create a managed IAM policy
+3. Define a managed IAM policy
 
     sample utapi policy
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,9 +109,26 @@ Server is running.
...
+
+    ```
+    "Resource": "arn:scality:utapi:::buckets/foo"
+    ```
+
+    Sample ARN limiting the user's access to metrics of an account with the ID
+    '012345678901':
+
+    ```
+    "Resource": "arn:scality:utapi::012345678901:buckets/*"
```
rahulreddy: Let's change the default policy to include accountId. And here we can refer to on how to access metrics for all accounts.

README.md
```diff
@@ -102,16 +102,69 @@ Server is running.
...
-    The user can also be restricted to only some buckets
-    (foo bucket for example) by changing the `Resource` property to
-    `"Resource": "arn:scality:utapi:::buckets/foo"`
+    In the above sample, the `Resource` property includes a series of Amazon
+    Resource Names (ARNs) used to define which resources the policy applies to.
+    Thus the sample policy applies to a user with an account ID '012345678901',
+    and grants access to metrics at the levels 'buckets', 'accounts', and
+    'users'.
+
+    The account ID of the ARN at any metric level can also be omitted. Extending
```
electrachong: If you omit it, does that mean any account can access those resources?

bennettbuchanan: Yes. Just updated to clarify.

README.md
```diff
@@ -102,16 +102,69 @@ Server is running.
...
-    The user can also be restricted to only some buckets
-    (foo bucket for example) by changing the `Resource` property to
-    `"Resource": "arn:scality:utapi:::buckets/foo"`
+    In the above sample, the `Resource` property includes a series of Amazon
+    Resource Names (ARNs) used to define which resources the policy applies to.
+    Thus the sample policy applies to a user with an account ID '012345678901',
+    and grants access to metrics at the levels 'buckets', 'accounts', and
+    'users'.
+
+    The account ID of the ARN at any metric level can also be omitted, meaning
```
electrachong: The account ID of the ARN ... can also be omitted, allowing any account to access metrics for those resources. As an example, we can extend the above sample policy to allow any account to access metrics at the level 'service':

bennettbuchanan: Sounds good 👍 . Updated.

PR #138, opened 2017-06-05
Refactor: Use generic methods array

bennettbuchanan:
Use a single array for any generic method. Use this array to build necessary object mappings and run associated tests.

This reduces the task of adding a new generic method to Utapi to just updating this array and the S3MetricResponse object.

Note that this also adds a missing test for the 'createBucket' operation in testMetrics.js, hence the additional 4 unit tests (one additional test for each of the four metric levels).


bennettbuchanan:
@ironman-machine try

rahulreddy:
Good work! I may have some cosmetic changes here and there.

bennettbuchanan:
I've been wanting to do that for a while 😺

philipyoo (approved):

bennettbuchanan:
These changes can be added if expanding Utapi metrics going forward.

PR #143, opened 2017-08-03
ft: propagate redis password from config

rahulreddy:


rahulreddy:
@ironman-machine try

rahulreddy:
@ironman-machine r+

PR #146, opened 2017-08-04
Fix builds relatively to scality/Arsenal#302

DavidPineauScality:


DavidPineauScality:
Merging manually, as this only impact the repo's builds.

PR #149, opened 2017-08-23
FT: Allow no auth for request

bennettbuchanan:
For open source usage or testing purposes, we can allow getting metrics without Vault.

### DIFF
--- a/lib/Vault.js
+++ b/lib/Vault.js
@@ -42,6 +42,9 @@ class Vault {
      * @return {undefined}
     */
     authenticateV4Request(params, requestContexts, callback) {
+        if (process.env.NO_AUTH) {
+            return callback(null, undefined);
+        }
         const { accessKey, signatureFromRequest, region, scopeDate,
             stringToSign }
             = params.data;


lib/Vault.js
```diff
@@ -42,6 +42,9 @@ class Vault {
      * @return {undefined}
     */
     authenticateV4Request(params, requestContexts, callback) {
+        if (process.env.NO_AUTH) {
```
rahulreddy: I would suggest setting up auth like S3 does, using an `authData.json` and we can use an env. var. like
`UTAPI_AUTH=MEMORY`.

JibranKalia:
@bennettbuchanan Is there any way I can help with this?

bennettbuchanan:
Hi @JibranKalia, thank you for the offer! I haven't had the chance to revisit this (I was intending to get it merged before our next hackathon). If you want to take over this branch, please do 😸. As @rahulreddy suggests, you might compare how S3 is using `authData.json` for in memory authentication and implement that here.

rahulreddy:
@JibranKalia  Thanks for the pitching in to manage this feature. IMO the correct way to do this is to have some sort of auth instead of no auth. My plan is to implement in-memory auth similar to how S3 implemented it. Since it involves traversing multiple modules, I will take it upon myself to make a PR implementing the feature. Plan is to target it this week.

PR #152, opened 2017-09-08
Publish actions with side effects

ploki:
When a method with side effect is performed, publish
a message in the bucket channel to notify subscribers
for state changes in the storage.

the channel is defined as ```s3:event:<bucket>```
the message is defined as ```<method>[:<object>[:<object>[:...]]]```

### DIFF
--- a/lib/UtapiClient.js
+++ b/lib/UtapiClient.js
@@ -6,42 +6,53 @@ const { errors } = require('arsenal');
 const redisClient = require('../utils/redisClient');
 
 const methods = {
-    createBucket: '_genericPushMetric',
-    deleteBucket: '_genericPushMetric',
-    listBucket: '_genericPushMetric',
-    getBucketAcl: '_genericPushMetric',
-    putBucketAcl: '_genericPushMetric',
-    putBucketCors: '_genericPushMetric',
-    getBucketCors: '_genericPushMetric',
-    deleteBucketCors: '_genericPushMetric',
-    putBucketWebsite: '_genericPushMetric',
-    getBucketWebsite: '_genericPushMetric',
-    getBucketLocation: '_genericPushMetric',
-    deleteBucketWebsite: '_genericPushMetric',
-    uploadPart: '_pushMetricUploadPart',
-    initiateMultipartUpload: '_genericPushMetric',
-    completeMultipartUpload: '_pushMetricCompleteMultipartUpload',
-    listMultipartUploads: '_pushMetricListBucketMultipartUploads',
-    listMultipartUploadParts: '_genericPushMetric',
-    abortMultipartUpload: '_genericPushMetricDeleteObject',
-    deleteObject: '_genericPushMetricDeleteObject',
-    multiObjectDelete: '_genericPushMetricDeleteObject',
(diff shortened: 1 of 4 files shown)


lib/UtapiClient.js
```diff
@@ -295,9 +327,14 @@ class UtapiClient {
         }
         const log = this.log.newRequestLoggerFromSerializedUids(reqUid);
         const timestamp = UtapiClient.getNormalizedTimestamp();
-        if (this[methods[metric]]) {
-            return this[methods[metric]](params, timestamp, metric, log,
-                callback);
+        if (methods[metric] && this[methods[metric].method]) {
+            return this[methods[metric].method](
+                params, timestamp, metric, log, () => {
```
bennettbuchanan: No check for error?

lib/UtapiClient.js
```diff
@@ -264,6 +275,40 @@ class UtapiClient {
...
+    * @param {object} log - logger
+    * @param {callback} [callback] - (optional) callback to call
+    * @return {undefined}
+    */
+    _publishEvent(metric, params, log, callback) {
+        const { bucket, objects } = params;
+        let message = metric;
+        if (Array.isArray(objects)) {
+            objects.forEach(object => {
+                message = `${message}:${object}`;
```
philipyoo: Do you mean to append here?

ploki: yes

rahulreddy: For readability, I recommend
```js
const message = keys.reduce((a, b) => `{a}:${b}`, metric);
```
Note, I renamed objects to keys.

lib/UtapiClient.js
```diff
@@ -264,6 +275,40 @@ class UtapiClient {
     }
 
     /**
+    * Publish an event in the bucket's dedicated channel to notify subscribers
+    * for performed actions with side effects
+    * @param {string} metric - metric to be published
+    * @param {object} params - params for the metrics
+    * @param {string} [params.bucket] - (optional) bucket name
+    * @param {array} [params.objects] - (optional) objects affected
```
rahulreddy: Since `objects` can be confused with JS objects, can you call this param keys instead? The doc can also reflect what type of elements the array contains. For example, if it's an array of strings then
```
* @param {string[]} [params.keys] - (optional) objects in a bucket that are affected
```

lib/UtapiClient.js
```diff
@@ -264,6 +275,40 @@ class UtapiClient {
...
+        }
+        return this.ds.publish(`s3:event:${bucket}`, message, err => {
+            if (err) {
+                log.error('Failed to publish event, ignored', {
+                    bucket,
+                    message,
+                    error: err,
+                });
+            }
+            if (typeof callback === 'function') {
```
rahulreddy: We already have this check in the calling function, so you can call the callback directly.

ploki: that's not the case with the command ```npm run ft_test```

rahulreddy:
@ironman-machine r+

rahulreddy:
@ironman-machine r- r+

rahulreddy:
Merging manually as the failures are unrelated.

PR #153, opened 2017-09-12
Chore: upgrade eslint

philipyoo:
List of Eslint rules: https://eslint.org/docs/rules/

Note: Will squash commits after making any changes based on peoples' feedback

### DIFF
--- a/examples/javascript-V4Auth-example.js
+++ b/examples/javascript-V4Auth-example.js
@@ -1,5 +1,5 @@
 const http = require('http');
-const aws4 = require('aws4');
+const aws4 = require('aws4'); // eslint-disable-line import/no-unresolved
 
 // Input AWS access key, secret key, and session token.
 const accessKeyId = 'EO4FRH6BA2L7FCK0EKVT';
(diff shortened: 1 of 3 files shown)


examples/javascript-V4Auth-example.js
```diff
@@ -1,5 +1,5 @@
 const http = require('http');
-const aws4 = require('aws4');
+const aws4 = require('aws4');  // eslint-disable-line
```
bennettbuchanan: Seems like maybe we should just disable linting of the examples directory if that's not difficult to do.

philipyoo: I like the idea :) Added changes to ignore the `examples` dir

rahulreddy: Seems to be a linter error
```
/home/ubuntu/utapi/examples/javascript-V4Auth-example.js
  2:22  error  Unable to resolve path to module 'aws4'  import/no-unresolved

```

rahulreddy:
@ironman-machine r+

philipyoo:
@ironman-machine try

rahulreddy:
@ironman-machine r- r+

PR #156, opened 2017-09-18
FIX: Return InternalError if state metric is negative

bennettbuchanan:


bennettbuchanan:
@ironman-machine
DEFAULT_BRANCH=rel/6.4
SCALITY_INTEGRATION_BRANCH=rel/6.4
try

src/lib/ListMetrics.js
```diff
@@ -257,6 +257,7 @@ export default class ListMetrics {
                     const m = getMetricFromKey(key);
                     let count = parseInt(item[1], 10);
                     count = Number.isNaN(count) ? 0 : count;
+                    count = count < 0 ? 0 : count;
```
rahulreddy: Since this affects only `storageUtilized` and `numberOfObjects`, can you add a comment saying so. It will be easy for maintenance.

src/lib/ListMetrics.js
```diff
@@ -237,6 +241,10 @@ export default class ListMetrics {
                 }
             });
 
+            if (isAnyMetricNegative) {
+                return cb(errors.InternalError.customizeDescription(
+                    'Redis server is not ready'));
```
rahulreddy: Instead of saying `Redis server is not ready`, providing more context would be beneficial. Something like
```
Utapi is in transient state for this time period as metrics are being collected. Please try again in a few minutes.
```

src/lib/ListMetrics.js
```diff
@@ -225,6 +226,9 @@ export default class ListMetrics {
                 } else {
                     let val = parseInt(item[1], 10);
                     val = isNaN(val) ? 0 : val;
+                    if (val < 0) {
+                        isAnyMetricNegative = true;
```
rahulreddy: You can use `Array.prototype.every` to exit early when any metric is negative.

bennettbuchanan:
@ironman-machine
try

src/lib/ListMetrics.js
bennettbuchanan: Updated.

src/lib/ListMetrics.js
bennettbuchanan: Updated.

bennettbuchanan:
@ironman-machine try

rahulreddy:
@ironman-machine r+

PR #158, opened 2017-09-19
FT: Add example to update sorted set metrics

bennettbuchanan:


examples/update-states.js
```diff
@@ -0,0 +1,103 @@
...
+}
+
+const redisClient = new Redis(Object.assign({
+    enableOfflineQueue: true,
+    keepAlive: 3000,
+}, REDIS_ENDPOINT));
+
+async.each(allParams, (params, cb) => {
+    const { storageUtilized, numberOfObjects } = params;
+    const timestamp = getCurrrentTimestamp();
```
philipyoo: extra `r` here

examples/update-states.js
```diff
@@ -0,0 +1,103 @@
...
+    const cmds = [];
+    if (storageUtilized !== undefined) {
+        const storageUtilizedKey = generateStateKey(params, 'storageUtilized');
+        cmds.push(
+            ['zremrangebyscore', storageUtilizedKey, timestamp, timestamp],
```
philipyoo: not familiar with redis so just making sure here, did you want a 2d-array here?

philipyoo (approved):

electrachong (approved):

examples/update-states.js
bennettbuchanan: Oops, just a typo! Updated.

examples/update-states.js
bennettbuchanan: This is how ioredis module does a batch of commands. See https://github.com/luin/ioredis.

For example:

```js
redis.multi([
  ['set', 'foo', 'bar'],
  ['get', 'foo']
]).exec(function () { /* ... */ });
```

examples/update-states.js
```diff
@@ -0,0 +1,103 @@
...
+
+const allStates = {
+    buckets: BUCKET_STATES,
+    accounts: ACCOUNT_STATES,
+    users: USER_STATES,
+    service: SERVICE_STATE,
+};
+
+const allParams = [];
+Object.keys(allStates).forEach(key => allStates[key].forEach(state =>
```
jonathan-gramain: For an example processing, I would think it's better to just declare the final `allParams` array statically instead of post-processing the various exploded arrays, even though there is some redundancy, because the reader will have less steps to understand what is actually provided to redis API. The user is likely to construct the redis commands his own way to match his inputs, which may not be similar to the structures here.

bennettbuchanan: Updated to use a single array.

examples/update-states.js
```diff
@@ -0,0 +1,103 @@
...
+    allParams.push(Object.assign(state, { service: 's3', level: key }))));
+
+function generateStateKey(params, metric) {
+    const { bucket, accountId, userId, level, service } = params;
+    const id = bucket || accountId || userId || service;
+    return `${service}:${level}:${id}:${metric}`;
+}
+
+function getCurrentTimestamp() {
+    const time = new Date(Date.now());
```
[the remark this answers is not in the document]
bennettbuchanan: True. Thanks!

examples/update-states.js
```diff
@@ -0,0 +1,89 @@
...
+        service: '<service-name>', // required
+        storageUtilized: '0',
+        numberOfObjects: '0',
+    },
+];
+
+function generateStateKey(params, metric) {
+    const { bucket, accountId, userId, service, resource } = params;
+    const id = bucket || accountId || userId || service;
+    return `s3:${resource}:${id}:${metric}`;
```
rahulreddy: Can you use the method in schema.js to generate this state key?

rahulreddy (changes requested):

examples/update-states.js
bennettbuchanan: I had done that originally, but thought it best to not be tied to any Utapi source code. That way users can take it and run it wherever they want without cloning Utapi, or using any particular Docker container.

rahulreddy: That makes sense, it is a good flexibility to have.

rahulreddy (approved):

PR #165, opened 2018-01-10
Fix/utapi versioning metrics

rahulreddy:


bennettbuchanan:
Closing this as this is no longer active. We can reopen if needed.

PR #171, opened 2018-03-26
FIX: Node engine

alexandre-merle:
Relax node engine to version superior to the current one, to allow upgrade


alexandre-merle:
@ThibaultRiviere done :)

PR #176, opened 2018-06-04
ft: get storage used by a resource

alexandre-merle:
Resource can be an account or user or bucket or service.

### DIFF
--- a/lib/UtapiClient.js
+++ b/lib/UtapiClient.js
@@ -898,6 +898,28 @@ class UtapiClient {
             return undefined;
         });
     }
+
+    /**
+     * Get storage used by bucket/account/user/service
+     * @param {object} params - params for the metrics
+     * @param {string} [params.bucket] - (optional) bucket name
+     * @param {string} [params.accountId] - (optional) account canonical ID
+     * @param {string} [params.userId] - (optional) account ID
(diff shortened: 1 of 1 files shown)


lib/UtapiClient.js
```diff
@@ -898,6 +898,28 @@ class UtapiClient {
...
+     * @param {string} [params.accountId] - (optional) account canonical ID
+     * @param {string} [params.userId] - (optional) account ID
+     * @param {string} [params.level] - (required) level of granularity
+     * @param {string} [params.service] - (required) service name (s3 for ex.)
+     * @param {object} log - Werelogs request logger
+     * @param {callback} callback - callback to call
+     * @return {undefined}
+     */
+    getStorageUtilized(params, log, callback) {
+        const key = generateCounter(params, 'storageUtilizedCounter');
```
DavidPineauScality: Is there no need for a "time window" for this request ? (I thought that everything was driven through time windows in utapi ?).
Otherwise code & doc look okay

rahulreddy: Since quotas need current usage, this responds with the global counter which represents the storage used at the moment.

bennettbuchanan:
Duplicate of https://github.com/scality/utapi/pull/188.

PR #177, opened 2018-06-05
ft: ZENKO 465 Utapi docker image

dora-korpar:
Turns Utapi into stand-alone service with Docker image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM node:6-slim
+
+WORKDIR /usr/src/app
+
+COPY package.json /usr/src/app
+
+RUN apt-get update \
+    && apt-get install -y jq --no-install-recommends \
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM node:8
```
rahulreddy: We are still unfortunately in node:6

index.js
```diff
@@ -1,6 +1,11 @@
 'use strict'; // eslint-disable-line strict
+const utapiServer = require('./lib/server');
+const config = require('./lib/Config');
+
+utapiServer(config);
```
rahulreddy: Since Utapi is included as a lib for S3, starting the server on index.js is not a good idea. You can use another file start_server.js or something instead.

PR #180, opened 2018-06-22
Ft: ZENKO 142 location quota metric

dora-korpar:
Adds functions for push/list of a location storage quota metric.

### DIFF
--- a/lib/UtapiClient.js
+++ b/lib/UtapiClient.js
@@ -59,6 +59,7 @@ const metricObj = {
     buckets: 'bucket',
     accounts: 'accountId',
     users: 'userId',
+    location: 'location',
 };
 
 class UtapiClient {
@@ -80,7 +81,7 @@ class UtapiClient {
         const api = (config || {}).logApi || werelogs;
         this.log = new api.Logger('UtapiClient');
         // By default, we push all resource types
-        this.metrics = ['buckets', 'accounts', 'users', 'service'];
+        this.metrics = ['buckets', 'accounts', 'users', 'service', 'location'];
         this.service = 's3';
         this.disableClient = true;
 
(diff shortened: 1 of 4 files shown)


lib/UtapiClient.js
```diff
@@ -898,6 +899,69 @@ class UtapiClient {
             return undefined;
         });
     }
+
+    /**
+     *
+     * @param {string} location - name of data location
+     * @param {number} updateSize - size of new object in location in bytes
```
jonathan-gramain: I think you also want to mention that it may be negative for deleted objects.

lib/schema.js
```diff
@@ -74,9 +74,13 @@ function getSchemaPrefix(params, timestamp) {
 */
 function generateKey(params, metric, timestamp) {
     const prefix = getSchemaPrefix(params, timestamp);
+    if (params.location) {
```
jonathan-gramain: Is this special case needed, what about simply adding the `locationStorage` key to the global `keys` object? It looks like what you added to support the location property in `getSchemaPrefix()` was meant for this.

bennettbuchanan: I agree that we should maintain the common keys objects, perhaps updating one of the "state keys" (or at least move this logic in `getSchemaPrefix`. My feeling is that the metrics should be exposed through an API listing as are other metrics that Utapi is tracking.

lib/UtapiClient.js
```diff
@@ -898,6 +899,70 @@ class UtapiClient {
...
+    }
+
+    /**
+     *
+     * @param {string} location - name of data backend to get metric for
+     * @param {string} reqUid - Request Unique Identifier
+     * @param {function} callback - callback to call
+     * @return {undefined}
+     */
+    getLocationMetric(location, reqUid, callback) {
```
bennettbuchanan: I would expect the metrics retrieval to be follow how Utapi lists other metrics (i.e., through a GET request to the API endpoint). This seems like a relatively big departure from the Utapi design.

lib/UtapiClient.js
```diff
@@ -898,6 +899,70 @@ class UtapiClient {
...
+    /**
+     *
+     * @param {string} location - name of data location
+     * @param {number} updateSize - size in bytes to update location metric by,
+     * could be negative, indicating deleted object
+     * @param {string} reqUid - Request Unique Identifier
+     * @param {function} callback - callback to call
+     * @return {undefined}
+     */
+    pushLocationMetric(location, updateSize, reqUid, callback) {
```
bennettbuchanan: Why not make this a private function that goes through the common `pushMetric`?

lib/UtapiClient.js
dora-korpar: We made the decision to keep the location metric (both push & get) separate from the rest of Utapi functionality. It's definitely a departure from the original design, but it will be used differently from the original push & list. The get is going to be used by S3 regularly, so it makes more sense to create a method rather than continually making HTTP requests.

dora-korpar:
Can't reply to the `getSchemaPrefix` comment directly, but I ultimately agree about the keys object. The conditional was included for now in the interest of time.

PR #208, opened 2019-04-29
feature: S3C-2144 Allow disabling authentication

bennettbuchanan:
Allow running the UTAPI server locally and without AWS Signature Version 4 authentication.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,14 @@ To run the server:
 npm start
 ```
 
+To run the server without using the [Signature Version 4 Signing Process](
+https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html), set the
+environment variable `NO_AUTH` to `'true'`:
+
+```
+NO_AUTH=true npm start
+```
+
 ## Client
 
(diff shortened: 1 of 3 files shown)


bennettbuchanan:
@bert-e help

bennettbuchanan:
@bert-e reset

rahulreddy (commented):
I would suggest using `NO_AUTH`. I am a bit iffy on disabling auth for tests as it creates a testing gap where auth is enabled. In that case the tests need to be run with and without auth which brings the point to is it worth doing this? It could be, I am curious what the motivation is.

bennettbuchanan:
@rahulreddy `NO_AUTH` sounds good to me.

We will still run all the authentication tests (the ones currently in Integration), but will use a Vault mock server when doing functional testing.

Mostly, the change will allow anyone to use UTAPI without running a Vault server, reducing the "barrier to entry".

bennettbuchanan:
Closing this PR because the change will be revisited and reopened once migrating UTAPI to the Zenko stack.

PR #215, opened 2019-04-30
bugfix: S3C-2155 Time range validation

bennettbuchanan:
I would like to eventually refactor this to allow for custom response messages based on the validation checks themselves. However, these checks will serve in the interim to prevent unexpected behavior related to time ranges.

Changes made:
* Check that both start and end times are in the past
* Do not allow the start time to be greater than the end time, especially in the case where the user does not provide an end time

### DIFF
--- a/src/validators/validateTimeRange.js
+++ b/src/validators/validateTimeRange.js
@@ -37,8 +37,13 @@ export default function validateTimeRange(timeRange) {
                 return false;
             }
         }
-
-        if (timeRange[0] > timeRange[1]) {
+        const now = Date.now();
+        // If end is not provided, it is later set as the current timestamp.
+        const endTime = timeRange[1] || now;
+        if (endTime > now) {
+            return false;
+        }
+        if (timeRange[0] > endTime) {
             return false;
         }
         return true;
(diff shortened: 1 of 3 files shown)


src/validators/validateTimeRange.js
```diff
@@ -37,8 +37,16 @@ export default function validateTimeRange(timeRange) {
                 return false;
             }
         }
-
-        if (timeRange[0] > timeRange[1]) {
+        const now = Date.now();
+        if (timeRange[0] > now) {
```
jonathan-gramain: I think this check is unnecessary as it is covered by the following checks.

jonathan-gramain (approved):

src/validators/validateTimeRange.js
bennettbuchanan: Thanks, I've removed it now 😸

tmacro (approved):

dora-korpar (approved):

rahulreddy (approved):

bennettbuchanan:
@bert-e help

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e approve

PR #232, opened 2019-05-08
bugfix: S3C-1506 Update start/end reducer

bennettbuchanan:
The values `storageUtilized` and `numberOfObjects` should be an array where the 0th element is the value of the first response, and the 1st element is the value of the last response, rather than a sum of values. I.e. the same logic as `timeRange`.


anurag4DSB (approved):

anurag4DSB (approved):

philipyoo (approved):

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e approve

PR #236, opened 2019-05-08
bugfix: S3C-2105 Add put data metric

bennettbuchanan:
Add a `putData` metric which corresponds to the route backbeat method `putData` which serves as the endpoint for replicated data operations.


anurag4DSB (approved):

dora-korpar (approved):

bennettbuchanan:
@bert-e approve

PR #242, opened 2019-05-09
bugfix: S3C-2155 Allow end time in the future

bennettbuchanan:
It is possible that some existing requests use an end time that is in the future (e.g. Integration tests use such a range). It's possible that we may want to revise the API specifications, perhaps when migrating UTAPI to the Zenko stack. However currently the risk of breaking compatibility is too great.


bennettbuchanan:
@bert-e reset

JianqinWang (approved):

philipyoo (approved):

bennettbuchanan:
@bert-e approve

bennettbuchanan:
@bert-e status

PR #259, opened 2019-05-28
bugfix: S3C-2076 Add UtapiReindex

bennettbuchanan:
Add the UTAPI reindex feature which queries bucketd for current storage utilized and number of object values, and updates Redis keys accordingly. The operation is performed at a regular interval defined by the user, and set on a default of once per week.

Thank you to @patrickdos for authoring the scripts and providing a solution for the out-of-sync metrics. For python script sources see https://bitbucket.org/scality-ts/utapi-reindex.

### DIFF
--- a/lib/reindex/reporting.py
+++ b/lib/reindex/reporting.py
@@ -1,0 +1,102 @@
+import requests
+import redis
+import json
+import ast
+import sys
+import time
+import urllib
+import re
+import sys
+from threading import Thread
+from concurrent.futures import ThreadPoolExecutor
+
+if len(sys.argv) == 6:
+    ip = sys.argv[1]
+    port = sys.argv[2]
+    sentinel_cluster_name = sys.argv[3]
+    bucketd_host = sys.argv[4]
+    bucketd_port = sys.argv[5]
+    print("Sentinel IP used: %s" % ip)
+    print("Sentinel port used: %s" % port)
+    print("Sentinel cluster name used: %s" % sentinel_cluster_name)
+    print("BucketD host used: %s" % bucketd_host)
+    print("BucketD port used: %s" % bucketd_port)
+else:
+    ip = "127.0.0.1"
+    port = "16379"
+    sentinel_cluster_name = "scality-s3"
+    bucketd_host =  "127.0.0.1"
+    bucketd_port = "9000"
+
+
+def safe_print(content):
+    print("{0}".format(content))
+
+
+class askRedis():
+
+    def __init__(self, ip="127.0.0.1", port="16379", sentinel_cluster_name="scality-s3"):
+
+        r = redis.Redis(host=ip, port=port, db=0)
+        self._ip, self._port = r.sentinel_get_master_addr_by_name(sentinel_cluster_name)
+
+    def read(self, resource, name):
+
+        r = redis.Redis(host=self._ip, port=self._port, db=0)
+        res = 's3:%s:%s:storageUtilized:counter' % (resource, name)
+        total_size = r.get(res)
+        res = 's3:%s:%s:numberOfObjects:counter' % (resource, name)
+        files = r.get(res)
+        return {'files': int(files), "total_size": int(total_size)}
+
+
+class S3ListBuckets():
+
+    def __init__(self, ip="127.0.0.1", bucketd_port="9000"):
+
+        self.ip = ip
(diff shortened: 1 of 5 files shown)


lib/reindex/reporting.py
```diff
@@ -0,0 +1,89 @@
...
+    print("Sentinel IP used %s" % ip)
+else:
+    ip = "127.0.0.1"
+
+
+def safe_print(content):
+    print("{0}".format(content))
+
+
+class askRedis():
```
philipyoo, line 36: TIL: use of parens for class definitions in python are a thing o.o

lib/reindex/reporting.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __init__(self, ip="127.0.0.1", port="16379"):
+
+        r = redis.Redis(host=ip, port=port, db=0)
+        self._ip, self._port = r.sentinel_get_master_addr_by_name('scality-s3')
+
+    def read(self, ressource, name):
```
philipyoo: minor: extra `s` in `ressource`, but its used in method as is so not a big deal

lib/reindex/s3_bucketd.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+class updateRedis():
+
+    def __init__(self, ip="127.0.0.1", port="16379"):
+
+        r = redis.Redis(host=ip, port=port, db=0)
+        self._ip, self._port = r.sentinel_get_master_addr_by_name('scality-s3')
+
+    def read(self, ressource, name):
```
philipyoo: minor: `ressource` extra `s`

lib/reindex/s3_bucketd.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def update(self, ressource, name, size, files):
+
+        timestamp = int(time.time() - 15 * 60) * 1000
+        r = redis.Redis(host=self._ip, port=self._port, db=0)
+
+        numberOfObjects = 's3:%s:%s:numberOfObjects' % (ressource, name)
+        storageUtilized = 's3:%s:%s:storageUtilized' % (ressource, name)
+
+        r.zremrangebyscore(numberOfObjects, timestamp, timestamp)
```
philipyoo, line 58: just curious, why are sorted sets used for these stats?

lib/reindex/s3_bucketd.py
```diff
@@ -0,0 +1,203 @@
...
+            Contents = payload["Contents"]
+            return (r.status_code, payload, Contents)
+        else:
+            return (r.status_code, "", "")
+
+    def retkeys(self, Contents):
+        total_size = 0
+        files = 0
+        key = ""
+        user = "Unknow"
```
philipyoo: minor: `Unknown`

lib/reindex/reporting.py
bennettbuchanan: I think it is unnecessary but harmless—the parens may be used for class inheritance. https://docs.python.org/3.4/tutorial/classes.html#inheritance.

lib/reindex/s3_bucketd.py
bennettbuchanan: Does this section in the Design doc help answer that?

https://github.com/scality/utapi/blob/development/8.0/DESIGN.md#sorted-sets

lib/reindex/reporting.py
philipyoo: yea, it was interesting to see, I guess introduced in python3

lib/reindex/s3_bucketd.py
philipyoo: nice ty!

bennettbuchanan:
@bert-e reset

philipyoo (approved):
Maybe the redis ports in python scripts may need to be passed, but overall lgtm 👍

bennettbuchanan:
Thanks, @philipyoo. Good point, and the other values are configurable as well (e.g. the sentinel cluster name), so these need to be updated from a config.

rahulreddy (approved):

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e approve

PR #274, opened 2019-06-05
bugfix: S3C-2076 Update default reindex schedule

bennettbuchanan:
The crontab is incorrect. It should, by default, run at midnight once per week (i.e. every Sunday).

Note the format is defined [here](https://www.npmjs.com/package/node-schedule#cron-style-scheduling), and differs slightly from the standard crontab format.


JianqinWang (approved):

philipyoo (approved):

bennettbuchanan:
@bert-e approve

PR #294, opened 2019-06-18
feature: S3C-2260 Add unit and functional testing

bennettbuchanan:
Maintenance testing for `UtapiReindex`:
* start Redis Sentinel as a process
* add basic tests for `UtapiReindex` methods connecting to Redis Sentinel

### DIFF
--- a/eve/workers/unit_and_feature_tests/redis/sentinel.conf
+++ b/eve/workers/unit_and_feature_tests/redis/sentinel.conf
@@ -1,0 +1,35 @@
+# Example sentinel.conf
+
+# The port that this sentinel instance will run on
+port 16379
+
+# Specify the log file name. Also the empty string can be used to force
+# Sentinel to log on the standard output. Note that if you use standard
+# output for logging but daemonize, logs will be sent to /dev/null
+logfile ""
+
+# dir <working-directory>
+# Every long running process should have a well-defined working directory.
+# For Redis Sentinel to chdir to /tmp at startup is the simplest thing
+# for the process to don't interfere with administrative tasks such as
+# unmounting filesystems.
+dir /tmp
+
+# sentinel monitor <master-name> <ip> <redis-port> <quorum>
+#
+# Tells Sentinel to monitor this master, and to consider it in O_DOWN
+# (Objectively Down) state only if at least <quorum> sentinels agree.
+#
+# Note that whatever is the ODOWN quorum, a Sentinel will require to
+# be elected by the majority of the known Sentinels in order to
+# start a failover, so no failover can be performed in minority.
+#
+# Replicas are auto-discovered, so you don't need to specify replicas in
+# any way. Sentinel itself will rewrite this configuration file adding
+# the replicas using additional configuration options.
+# Also note that the configuration file is rewritten when a
+# replica is promoted to master.
+#
+# Note: master name should not include special characters or spaces.
+# The valid charset is A-z 0-9 and the three characters ".-_".
+sentinel monitor scality-s3 127.0.0.1 6379 1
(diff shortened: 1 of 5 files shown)


eve/workers/unit_and_feature_tests/redis/sentinel.conf
```diff
@@ -0,0 +1,35 @@
...
+#
+# Replicas are auto-discovered, so you don't need to specify replicas in
+# any way. Sentinel itself will rewrite this configuration file adding
+# the replicas using additional configuration options.
+# Also note that the configuration file is rewritten when a
+# replica is promoted to master.
+#
+# Note: master name should not include special characters or spaces.
+# The valid charset is A-z 0-9 and the three characters ".-_".
+sentinel monitor scality-s3 127.0.0.1 6379 2
```
ssalaues: It's very likely this is silently failing. You have a quorum of `2` set but there is only 1 Redis server.

bennettbuchanan: I agree this should be updated to 1 to be accurate and have pushed that change.

However, if the sentinel can't connect to the Redis server the test will not fail silently regardless of the quorum number. The refused connect will fail with an error:

```
Error: connect ECONNREFUSED 127.0.0.1:6379
```

bennettbuchanan: Thanks, I'm looking into this a bit further to be certain. It's possible that the error does not come from the sentinel though.

ssalaues: Hmm that's interesting. Just let me know you need help debugging the environment. Sometimes peaking through the Berte/Eve curtains can be very useful.

bennettbuchanan: I just double checked this, and yes, the sentinel will respond with an error regardless of the quorum number when the client emits the 'error' event [here](https://github.com/scality/utapi/blob/development/8.0/lib/UtapiReindex.js#L165):

```
{ Error: connect ECONNREFUSED 127.0.0.1:6379
    at Object._errnoException (util.js:1026:11)
    at _exceptionWithHostPort (util.js:1049:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1174:14)
  code: 'ECONNREFUSED',
  errno: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 6379 }
```

Additionally, if the Redis Server is running but there is no Redis Sentinel, the response is as expected:

```
Error: All sentinels are unreachable. Retrying from scratch after 10ms. Last error: Connection is closed.
```

The client which is connecting to the Redis Server directly in the `beforeEach` [here](https://github.com/scality/utapi/pull/294/files#diff-474cc7d7664919efb3ec1d789deb7e21R13) will also emit the same error which is why I thought there was a chance it wasn't coming from the sentinel in fact.

In any case, setting the value to 1 seems more correct to me. I'll let you know if there is a need for debugging the environment. Thanks! 😃

tests/functional/testReindex.js
```diff
@@ -0,0 +1,64 @@
...
+
+    beforeEach(done => {
+        reindex = new UtapiReindex();
+        redis = redisClient({}, log)
+            .on('ready', done)
+            .on('error', done);
+    });
+
+    afterEach(done => {
+        redis.flushdb(done);
```
JianqinWang: maybe a good idea to close the client as well if we're re-instantiating a redisclient for every test

JianqinWang (approved):

ssalaues (approved):

tests/functional/testReindex.js
bennettbuchanan: Good idea, I've updated that.

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e approve

PR #327, opened 2019-07-18
bugfix: S3C-2317 Append UUID to sorted set members

bennettbuchanan:
A storageUtilized and/or numberOfObjects data point is lost when a sorted set member is added as the same value of an existing member in the sorted set.

This change appends a UUID to each sorted set member to guarantee uniqueness across the sorted set and thus preserve previous interval metrics.

### DIFF
--- a/eve/main.yml
+++ b/eve/main.yml
@@ -41,5 +41,14 @@ stages:
           name: run unit tests
           command: npm test
       - ShellCommand:
-          name: run feature tests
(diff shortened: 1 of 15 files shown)


alexanderchan-scality (approved):

alexanderchan-scality (approved):

jonathan-gramain (approved):

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e approve

PR #341, opened 2019-07-23
bugfix: S3C-2315 Support versioning with reindex

bennettbuchanan:
The reindex script is currently only listing master versions. This changes the script to list all versions for calculating the numberOfObjects and storageUtilized metrics.

### DIFF
--- a/lib/reindex/s3_bucketd.py
+++ b/lib/reindex/s3_bucketd.py
@@ -105,20 +105,21 @@ class S3BucketD(Thread):
         self.total_size = 0
         if self.mpu:
             self.bucket = "mpuShadowBucket"+bucket
-        self.seekeys = seekeys
         self.ip = ip
         self.bucketd_port = bucketd_port;
 
-    def listbucket(self, session=None, marker=""):
+    def listbucket(self, session=None, marker="", versionmarker=""):
         m = marker.encode('utf8')
         mark = urllib.parse.quote_plus(m)
-        params = "%s?listingType=DelimiterMaster&maxKeys=1000&marker=%s" % (
-            self.bucket, mark)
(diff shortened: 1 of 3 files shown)


tmacro (approved):

lib/reindex/s3_bucketd.py
```diff
@@ -105,20 +105,22 @@ def __init__(self, userid=None, bucket=None, mpu=False, seekeys=False, ip="127.0
...
 
-    def listbucket(self, session=None, marker=""):
+    def listbucket(self, session=None, marker="", versionmarker=""):
         m = marker.encode('utf8')
         mark = urllib.parse.quote_plus(m)
-        params = "%s?listingType=DelimiterMaster&maxKeys=1000&marker=%s" % (
-            self.bucket, mark)
+        v = versionmarker.encode('utf8')
+        versionmarker = urllib.parse.quote_plus(v)
+        extra = "&versionIdMarker=%s" % versionmarker
```
jonathan-gramain: From a style perspective, it feels odd to pass `versionIdMarker` as an extra param, it should be at the same level than `keyMarker` since it has the same importance. I suggest to add `&versionIdMarker...` to the format string and pass `versionmarker` in the format params.

anurag4DSB (approved):

jonathan-gramain (approved):

lib/reindex/s3_bucketd.py
bennettbuchanan: Good call, I've updated to that style and pushed the change.

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e approve

PR #349, opened 2019-07-25
bugfix: S3C-2317 Add uuid module as a dependency

bennettbuchanan:
The `uuid` module is already installed by another dependency. However UTAPI requires this module so it is best practice to add it as a dependency in the `package.json`.


JianqinWang (approved):

tmacro (approved):

bennettbuchanan:
@bert-e approve

nicolas2bert (approved):

PR #352, opened 2019-07-25
bugfix: S3C-2342 BucketD listing functional tests

bennettbuchanan:
Refactor functional tests to allow for a bucketD listing greater than max keys query param of 1000 and add a test for listings greater than that.

### DIFF
--- a/tests/functional/cron/testReindex.js
+++ b/tests/functional/cron/testReindex.js
@@ -183,14 +183,21 @@ describe('UtapiReindex', () => {
         });
     });
 
-    describe('::_scheduleJob', () => {
-        function waitUntilLockHasValue({ value }, cb) {
+    describe('::_scheduleJob', function test() {
+        this.timeout(30000);
+
+        function waitUntilLockHasValue({ value, job }, cb) {
             let shouldLeave;
+            let shouldCallJob = job !== undefined;
 
             async.doUntil(next =>
                 redis.get(REINDEX_LOCK_KEY, (err, res) => {
                     if (err) {
                         return next(err);
+                    }
+                    if (shouldCallJob) {
+                        job();
+                        shouldCallJob = false;
                     }
                     shouldLeave = res === value;
                     return setTimeout(next, 200);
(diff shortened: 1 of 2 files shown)


JianqinWang (approved):

tests/functional/cron/testReindex.js
```diff
@@ -213,60 +215,82 @@ describe('UtapiReindex', () => {
...
+                        contentLength: 1024,
+                    })
+                    .setBucketCount(count)
+                    .createBuckets();
+
+                    reindex._scheduleJob();
+
+                    // Wait until the scripts have finished reindexing.
+                    async.series([
+                        next => waitUntilLockHasValue({ value: 'true' }, next),
```
jonathan-gramain: Might be racy here if the reindex script finishes before the first check for `value: 'true'` that could lead to the test being stuck here (although maybe unlikely in practice)

tests/functional/cron/testReindex.js
```diff
@@ -213,60 +215,82 @@ describe('UtapiReindex', () => {
...
-                next => {
-                    const params = {
-                        resource: {
-                            type: 'buckets',
-                            buckets: [mock.values.BUCKET_NAME],
+        const bucketCounts = [1, 1001];
+
+        bucketCounts.forEach(count => {
+            describe(`bucket listing with a length of ${count}`, () => {
+                const bucket = `${mock.values.BUCKET_NAME}-${count - 1}`;
```
jonathan-gramain: why `count - 1` not `count`?

jonathan-gramain (approved):

tests/functional/cron/testReindex.js
bennettbuchanan: Because the bucket creation listing is zero-indexed [here](https://github.com/scality/utapi/pull/352/files#diff-6f8c535ed582ceee20e2ec4e8216539dR47). Sounds like it would be more intuitive to actually set `count` in the test and `+ 1` in the bucket creation. Will update to make it more intuitive.

tests/functional/cron/testReindex.js
bennettbuchanan: Good point. There is definitely potential for that here however unlikely because the single bucket listing test takes hundreds of MS. Good to prevent it from ever occurring though, so I'll update such that at least one call to Redis is made before the job is run.

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e approve

PR #357, opened 2019-07-31
improvement S3C-234 Operation counters config

bennettbuchanan:
Some customers may only need the metrics of `storageUtilized` and `numberOfObjects` for billing purposes. In such cases, the user can save significant space in the Redis DB and reduce the rate of its growth by limiting the metrics stored to just those.

To that end this PR introduces a feature that allows the user to not store S3 operation counter metrics (i.e. CreateBucket, PutObject, GetObject, etc.) as they are not needed for the computation of `storageUtilized` or `numberOfObjects`. Alternatively, the user may specify a select list of counters they want to track, and disable all others.

This feature is made available as configuration properties of `UtapiClient`. All operation counters are enabled by default (i.e. the original behavior). To disable all operation counters, the user may set `disableOperationCounters` to `true` in the configuration. Alternatively, if specifying `enabledOperationCounters`, only those operation counter will be enabled at the exclusion of all others.

### DIFF
--- a/lib/UtapiClient.js
+++ b/lib/UtapiClient.js
@@ -89,6 +89,8 @@ class UtapiClient {
         // By default, we push all resource types
         this.metrics = ['buckets', 'accounts', 'users', 'service'];
         this.service = 's3';
+        this.disableOperationCounters = false;
+        this.enabledOperationCounters = [];
         this.disableClient = true;
 
         if (config) {
@@ -112,6 +114,12 @@ class UtapiClient {
                 // The configuration uses the property `component`, while
                 // internally this is known as a metric level `service`.
                 this.service = config.component;
+            }
+            if (config.disableOperationCounters) {
+                this.disableOperationCounters = config.disableOperationCounters;
+            }
+            if (config.enabledOperationCounters) {
+                this.enabledOperationCounters = config.enabledOperationCounters;
             }
             this.disableClient = false;
             this.expireMetrics = config.expireMetrics;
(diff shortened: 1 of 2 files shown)


lib/UtapiClient.js
```diff
@@ -145,6 +153,14 @@ class UtapiClient {
         return this;
     }
 
+    _isCounterEnabled(action) {
+        if (this.enabledOperationCounters.length > 0) {
```
JianqinWang, line 157: if `this.disabledOperationCounters` is set to true while the list for `this.enabledOperationCounters` is not empty, do we defer to the list of enabled op counters and ignore the `disableOperationCounters` setting?

bennettbuchanan: Yes, that's the idea 😄. I figured if anyone were setting `enabledOperationCounters` then they'd likely want that to take precedence. It's worth clarifying though, and I'll add a check in the CloudServer config to apply a constraint to make sure the user gets what they expect.

JianqinWang: cool, everything else lgtm :)

JianqinWang (approved):

dora-korpar (approved):

tmacro (approved):

bennettbuchanan:
@bert-e status

bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e approve

PR #375, opened 2019-09-30
bugfix: S3C-2408 MPU overwrite

bennettbuchanan:
**Overview**
When a MPU object overwrites another object in a non-versioned bucket, the storage utilized and number of objects is incorrect.

**Cause**
The storage utilized values are incremented as parts are uploaded, and not during the complete MPU. Because no storage utilized metrics are updated during a complete MPU operation, an overwritten object is not accounted for.

**Proposed Solution**
When a complete MPU operation succeeds and it is overwriting an object, pass `oldByteLength` as a parameter and decrement the storage utilized value accordingly. Additionally, instead of incrementing the number of objects count, do not modify the value.

### DIFF
--- a/lib/UtapiClient.js
+++ b/lib/UtapiClient.js
@@ -634,6 +634,57 @@ class UtapiClient {
         });
     }
 
+    _multipartUploadOverwrite(params, timestamp, action, log, cb) {
+        const counterCommands = [];
+        const levels = this._getParamsArr(params);
+
+        levels.forEach(level => {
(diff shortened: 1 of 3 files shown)


bennettbuchanan:
@bert-e reset

bennettbuchanan:
@bert-e reset

rahulreddy:
@bert-e reset

tmacro (approved):

rahulreddy (approved):

PR #383, opened 2020-02-11
bugfix: S3C-2604-list-multiple-bucket-metrics

anurag4DSB:
**What does this PR do, and why do we need it?**
> Combined the _specificResource when the type of resource requested is 'bucket'. So instead of having multiple requestContexts we will have all the requests combined into one.
> Post NodeJS upgrade we could not query more than 7 buckets as in the new node version they introduced a limit of 8KB for HTTP packet parsing, commit in Node.js for reference  https://github.com/nodejs/node/commit/a8532d4d23.
> Customer reply on metrics supplied by UTAPI to charge their customers

**Which issue does this PR fix?**
fixes #[S3C-2604](https://scality.atlassian.net/browse/S3C-2604)

**Special notes for your reviewers**:
This is a quick workaround, we are gonna re-architecture UTAPI eventually.

### DIFF
--- a/router/Router.js
+++ b/router/Router.js
@@ -213,13 +213,12 @@ class Router {
         const validator = utapiRequest.getValidator();
         // specific resources will be names of buckets, accounts or users
         const specificResources = validator.get(resourceType);
-
-        const requestContexts = specificResources.map(specificResource =>
+        const requestContext = [];
+        requestContext.push(
             new policies.RequestContext(utapiRequest.getRequestHeaders(),
-            utapiRequest.getRequestQuery(), resourceType, specificResource,
-            utapiRequest.getRequesterIp(), utapiRequest.getSslEnabled(),
-            utapiRequest.getAction(), 'utapi')
-        );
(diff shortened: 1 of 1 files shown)


router/Router.js
```diff
@@ -213,13 +213,12 @@ class Router {
         const validator = utapiRequest.getValidator();
         // specific resources will be names of buckets, accounts or users
         const specificResources = validator.get(resourceType);
-
-        const requestContexts = specificResources.map(specificResource =>
+        let requestContexts = [];
```
rahulreddy: You can still use `const` here as you are not replacing the content of the variable. Another minor change is to rename it to singular `requestContext` since it is only one request context

rahulreddy (changes requested):

anurag4DSB:
@bert-e reset

router/Router.js
anurag4DSB: Good catch.

anurag4DSB:
@bert-e approve

rahulreddy (approved):

tmacro (approved):

PR #440, opened 2020-04-28
ZENKO-2503: update dockerfile

alexanderchan-scality:
- [ ] eve step: image upload

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,13 @@
-FROM node:6-slim
+FROM node:10.20.1-slim
 
 WORKDIR /usr/src/app
 
 COPY package.json /usr/src/app
 
 RUN apt-get update \
-    && apt-get install -y jq --no-install-recommends \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,11 +1,11 @@
-FROM node:6-slim
+FROM node:10.20.1-slim
 
 WORKDIR /usr/src/app
 
 COPY package.json /usr/src/app
 
 RUN apt-get update \
-    && apt-get install -y jq --no-install-recommends \
+    && apt-get install -y jq python git build-essential --no-install-recommends \
```
NicolasT, line 8: Do you really need `python` and `build-essential` at *runtime*, or only at 'build' time (and could they as such be removed within this very same `RUN` statement)?

alexanderchan-scality: `python` and `build-essential` are only required at `build` time; they can be removed once build has completed. will fix

rahulreddy (approved):

PR #441, opened 2020-05-14
ft(S3C-2878): Add Warp10 Dockerfile

tmacro:
First PR of V2! 🎉
Remove most of the v1 code in the first commit, leaving only a skeleton and some functional tests
Comment eve config to get the tests to pass
Add Warp10 Dockerfile
Minor Readme edits

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,4 @@
-dist
 node_modules/
 **/node_modules/
-logs
-*.log
-dump.rdb
+.git
+eve/
--- a/images/warp10/Dockerfile
+++ b/images/warp10/Dockerfile
@@ -1,0 +1,23 @@
+FROM warp10io/warp10:2.5.0-ci
+
+ENV S6_VERSION 2.0.0.1
+
+ENV WARP10_VERSION 2.5.0
+
+ENV WARP10_CONF_TEMPLATES ${WARP10_HOME}/conf.templates/standalone
+RUN wget https://github.com/just-containers/s6-overlay/releases/download/v${S6_VERSION}/s6-overlay-amd64.tar.gz -O /tmp/s6-overlay-amd64.tar.gz \
+    && tar xzf /tmp/s6-overlay-amd64.tar.gz -C / \
+    && rm -rf /tmp/s6-overlay-amd64.tar.gz
+
+# Modify Warp 10 default config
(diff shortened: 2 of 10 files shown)


NicolasT:
Instead of removing all files in the repo, since this is a 'new project', wouldn't it make sense to start a new tree in the repository (as we did for MetalK8s 2.0)?

anurag4DSB (approved):

tmacro:
@NicolasT That would make sense, if bert-e can be made to cooperate, or if we can guarantee that no change to 1.x will be made

alexanderchan-scality (approved):

images/utapi/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM node:12-alpine
```
anurag4DSB, line 1: Can we move to Node 12?

I support using node12 due to its nice new features. But I was told that we are going to maintain all components on v10 LT for now.

May be, @rahulreddy can confirm.

images/utapi/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+WORKDIR /usr/src/app
+
+COPY ./package.json ./yarn.lock /usr/src/app/
+
+RUN apk add git \
+            build-base \
+            openssl \
+            python && \
+    yarn install --production --frozen-lockfile && \
```
anurag4DSB: I would suggest adding a .yarnrc file for --frozen-lcokfile:
` --install.frozen-lockfile true`

And also, would be a good idea to replace the `--production` flag as an environment variable so that the dockerfile can be reused.
`ARG NODE_ENV=production`

images/utapi/Dockerfile
anurag4DSB: Can we move to Node 12?

I support using node12 due to its nice new features. But I was told that we are going to maintain all components on v10 LTS for now.

May be, @rahulreddy  can confirm.

anurag4DSB (changes requested):

images/utapi/Dockerfile
tmacro: Good suggestion, The utapi dockerfile slipped in by accident though this PR should really only be warp 10 stuff

images/utapi/Dockerfile
tmacro: I went with 12 because 10 EOL's on 05/2021. I don't think starting a new project using 10 makes sense as we'd have to upgrade in about a year anyway

tmacro:
@bert-e reset

anurag4DSB (approved):

tmacro:
@bert-e approve

PR #456, opened 2020-06-09
bf(S3C-3043): Bump scality/Guidelines to HEAD

tmacro:


miniscruff (approved):

lib/ListMetrics.js
```diff
@@ -157,12 +155,13 @@ class ListMetrics {
         const end = range[1] || Date.now();
         const subRangesCount = Math.floor((end - start) / MAX_RANGE_MS) + 1;
         const subRanges = [];
+        // eslint-disable-next-line no-plusplus
```
miniscruff, line 158: is it expecting += 1 here?

tests/utils/utils.js
```diff
@@ -38,6 +39,7 @@ function getAllResourceTypeKeys() {
         return counters.concat(keys);
     });
     // Concatenate each array of resourceType keys into one single array.
+    // eslint-disable-next-line prefer-spread
     return [].concat.apply([], allResourceTypeKeys);
```
miniscruff, line 43: I think this wants `return [...allResourceTypeKeys];`? Instead of just disabling eslint, but if the intent is to disable all new eslint issues and come back later that is fine.

miniscruff (approved):
I am a fan of the new eslint rules but it makes sense to disable them and come back. Updating like 8 major versions up is bound to complain a bit lol.

tests/utils/utils.js
tmacro: There's apparently more going on here as changing this causes a test to break. I have no idea how the final array is supposed to look.

anurag4DSB (approved):

anurag4DSB (approved):

PR #472, opened 2020-06-11
ft(S3C-2999): add redis cache client

tmacro:


tmacro:
@bert-e wait

tmacro:
@bert-e wait

miniscruff:
This code doesn't look like any of the other PR's I have reviewed, not that I mind as it looks very clean just wanted to point it out.
Also no tests anywhere.

tmacro:
@miniscruff I'm writing tests as we speak, thought I'd start gathering reviews in parallel

miniscruff:
> @miniscruff I'm writing tests as we speak, thought I'd start gathering reviews in parallel

ok, my only question so far really is all the async stuff. Not sure every method needs to be async but maybe it does idk. Also TDD is a wonderful thing :)

libV2/cache/backend/memory.js
```diff
@@ -0,0 +1,45 @@
+class MemoryCache {
+    constructor() {
+        this._data = {};
+        this._shards = {};
+    }
+
+    async getKey(key) {
```
naren-rajendran, line 20: Does it have to be `async`? I see multiple `async` functions here without `await`, any specific reason, not an issue though!

naren-rajendran:
> > @miniscruff I'm writing tests as we speak, thought I'd start gathering reviews in parallel
> 
> ok, my only question so far really is all the async stuff. Not sure every method needs to be async but maybe it does idk. Also TDD is a wonderful thing :)

I almost missed @miniscruff 's comment.

libV2/cache/backend/memory.js
tmacro: `async` here is mostly for the convenience. Since both memory and redis backends need the same API so they can be swapped out, memory needs needs to return a promise as well. Using `async` allows me to wrap the return value in a resolved promise automatically

tmacro:
re `async` most of the other `async` functions are just marked as such to enforce the returning of a Promise and to make it clear to a developer that the function return a promise

libV2/cache/backend/memory.js
naren-rajendran: Fair enough 👍

libV2/cache/backend/redis.js
```diff
@@ -0,0 +1,164 @@
...
+
+    async addToShard(shard, metric) {
+        const metricKey = schema.getUtapiMetricKey(this._prefix, metric);
+        const shardKey = schema.getShardKey(this._prefix, shard);
+        const logger = moduleLogger.with({ method: 'addToShard' });
+        logger.debug('adding metric to shard', { metricKey, shardKey });
+
+        let results;
+        try {
+            results = await this._redis
```
miniscruff: could avoid indexing later by doing...
`let [setResults, saddResults] = await ...`

miniscruff: Would also remove the need of the comments later as its more semantic

tests/utils/v2Data.js
```diff
@@ -0,0 +1,27 @@
+const uuid = require('uuid');
+const { UtapiMetric } = require('../../libV2/models');
+
+function range(n, step) {
+    const vals = [...Array(n).keys()];
+    if (step) {
+        return vals.map(i => i * step);
+    }
+    return vals;
+}
```
miniscruff, line 10: For efficiency reasons, I think a generator will work best for a range method ( copies python basically ).
```suggestion
function* range(count, step=1) {
    for (let i = 0; i < count; i+=step) {
      yield i;
    }
}
```

miniscruff: Oh if you meant for non-integers then I think this would work.
```suggestion
function* range(count, step=1) {
    let i = 0;
    while (i < count * step) {
      yield i;
      i += step;
    }
}
```

tests/utils/v2Data.js
```diff
@@ -0,0 +1,27 @@
...
+        timestamp,
+        uuid: uuid.v4(),
+    });
+}
+
+function generateFakeEvents(start, stop, count) {
+    const duration = stop - start;
+    const eventsEvery = duration / count;
+    return range(count, eventsEvery).map(i => makeEvent(Math.floor(start + i)));
+}
```
miniscruff, line 23: Same as above, a generator here would be nice.
```suggestion
function* generateFakeEvents(start, stop, count) {
    const duration = stop - start;
    const eventsEvery = duration / count;
    for (let time of range(count, eventsEvery)) {
      yield makeEvent(Math.floor(start + time));
    }
}
```

tests/functional/v2/cache/testRedisBackend.js
```diff
@@ -0,0 +1,90 @@
...
+        const expected = testValues.map(e =>
+            schema.getUtapiMetricKey(prefix, e));
+        return Promise.all(
+            testValues.map(event => cache.addToShard(shard, event)),
+        )
+            .then(() => cache.getKeysInShard(shard))
+            .then(keys => {
+                assert.deepStrictEqual(keys.sort(), expected.sort());
+            });
+    });
```
miniscruff: I just want to demo what this would look like using async + await, I know it is still up to discussion but this is a good example. Note I have not run the tests so a couple of tweaks might be in order but if serious thought is taken on converting it should not be too hard.

```suggestion
    it('should fetch the keys in a shard', async () => {
        const shard = shardFromTimestamp(new Date().getTime());
        const expected = testValues.map(e =>
            schema.getUtapiMetricKey(prefix, e)
        );
        await Promise.all(testValues.map(event => cache.addToShard(shard, event)));
        let keys = await cache.getKeysInShard(shard);
        assert.deepStrictEqual(keys.sort(), expected.sort());
    });
```

cc @rahulreddy @anurag4DSB

tests/functional/v2/cache/testRedisBackend.js
```diff
@@ -0,0 +1,90 @@
...
+                .then(value => {
+                    assert.deepStrictEqual(JSON.parse(value), event.getValue());
+                });
+        }));
+
+    testValues.map(event =>
+        it('should add a key to a shard', () => {
+            const shard = shardFromTimestamp(event.timestamp);
+            return cache.addToShard(shard, event).then(res => {
+                assert.strictEqual(true, res);
```
miniscruff: not a huge deal but strict equal has parameters of (value, expected) so when reading a failed test log it would say something like. "expected true to equal something" when it might make more sense the other way around. It doesn't really matter much when comparing a bool though.

tests/functional/v2/cache/testRedisBackend.js
```diff
@@ -0,0 +1,90 @@
...
+const startTime = new Date().getTime();
+const endTime = startTime + 30000; // Add 30 seconds;
+const testValues = generateFakeEvents(startTime, endTime, 50);
+
+describe('Test cache redis backend', () => {
+    let cache;
+    let prefix;
+    beforeEach(done => {
+        prefix = uuid.v4();
+        cache = new RedisCache({}, prefix);
```
miniscruff, line 19: should there be an afterEach to disconnect from redis?

libV2/models/Base.js
```diff
@@ -0,0 +1,75 @@
+const Joi = require('@hapi/joi');
+
+class BaseModel {
+    constructor(data) {
+        if (data !== undefined) {
+            this._data = data;
+        } else {
+            this._data = {};
+        }
```
miniscruff: Can just use `|| {}` here and it should work the same.
```suggestion
        this._data = data || {};
```

tests/unit/v2/cache/testMemoryBackend.js
```diff
@@ -0,0 +1,118 @@
...
+            });
+    });
+
+    it('should get all shards', () => Promise.all(
+        testValues.map(event => {
+            const shard = shardFromTimestamp(event.timestamp);
+            return mem.addToShard(shard, event.uuid);
+        }),
+    )
+        .then(() => mem.getShards())
```
miniscruff: The end parenthesis for promise.all makes it look like the it test is closed way before it actually is. Had me confused for a good few minutes.

miniscruff (changes requested):

tests/functional/v2/cache/testRedisBackend.js
tmacro: I'll double check but I believe passing an `async` function as a callback is a linting error

tmacro: I was wrong, thinking of this rule https://eslint.org/docs/rules/no-async-promise-executor

tests/utils/v2Data.js
tmacro: I agree that generators and iterators are nice, but they're currently banned by our use of `eslint-config-airbnb`. See https://github.com/airbnb/javascript#iterators--nope

miniscruff: dang, however much I think functional is great and all generators are just so powerful and efficient I would easily swap over to for..of over a forEach given. Airbnb's reasoning seems to be mostly related to the web and node is probably fine but I couldn't really get much info on it.

miniscruff (approved):
Approving even though I think generators are the way to go :thinking:

libV2/cache/backend/memory.js
```diff
@@ -0,0 +1,71 @@
...
+        return true;
+    }
+
+    async getKey(key) {
+        return this._data[key];
+    }
+
+    async setKey(key, data) {
+        this._data[key] = data;
+        return true;
```
jonathan-gramain, line 26: Unless you plan to change the return value to a more meaningful one, you could return the instance itself to be able to chain the calls to set multiple keys or do other stuff: `setKey(...).setKey(...)`

libV2/cache/backend/memory.js
```diff
@@ -0,0 +1,71 @@
...
+        if (this._shards[shard]) {
+            this._shards[shard].push(metricKey);
+        } else {
+            this._shards[shard] = [metricKey];
+        }
+        return true;
+    }
+
+    async getKeysInShard(shard) {
+        if (this._shards[shard]) {
```
jonathan-gramain: You may use the more compact `return this._shards[shard] || []`, but up to you

libV2/cache/backend/memory.js
```diff
@@ -0,0 +1,71 @@
...
+        delete this._shards[shard];
+        return true;
+    }
+
+    async getShards() {
+        return Object.keys(this._shards);
+    }
+
+    async shardExists(shard) {
+        return Object.keys(this._shards).includes(shard.toString());
```
jonathan-gramain: Unless it does not work for some reason, it could be more efficient and simple as `return this._shards[shard.toString()] !== undefined`

libV2/utils/log.js
```diff
@@ -0,0 +1,81 @@
...
+rootLogger.debug('logger initialized', { loggerConfig });
+
+function buildRequestLogger(req) {
+    let reqUids = [];
+    if (req.headers['x-scal-request-uids'] !== undefined) {
+        reqUids.push(req.headers['x-scal-request-uids']);
+    }
+
+    // Remove any overly long request ids
+    reqUids = reqUids.filter(id => id.length < 128);
```
jonathan-gramain: Instead of removing, I'd rather truncate them if length is an issue to keep some of the first request ID components.

libV2/utils/shard.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Returns a unix style timestamp floored to 10 second resolution
+ * @param {Number} timestamp - Unix timestamp with millisecond resolution
+ * @returns {Number} - Unix timestamp representing beginning of shard
+ */
+function shardFromTimestamp(timestamp) {
+    // Do this with string manipulation to avoid float errors
```
jonathan-gramain: You can also use Math functions like `Math.floor()` that are meant to deal with int/float conversions. String manipulation works though, just could be slightly cleaner.

tests/functional/v2/cache/testRedisBackend.js
```diff
@@ -0,0 +1,76 @@
...
+
+    describe('Test shardExists', () => {
+        it('should return false if the shard does not exists', async () => {
+            const shard = shardFromTimestamp(new Date().getTime() + 10000);
+            const res = await cache.shardExists(shard);
+            assert.strictEqual(res, false);
+        });
+
+        it('should return true if the shard exists', async () => {
+            const shard = shardFromTimestamp(new Date().getTime());
```
jonathan-gramain, line 70: Is there any risk of flakiness by specifying the current time that falls close to a shard boundary? If so, could be better to set a past time maybe

tests/unit/v2/cache/testMemoryBackend.js
```diff
@@ -0,0 +1,106 @@
...
+const { shardFromTimestamp } = require('../../../../libV2/utils');
+const schema = require('../../../../libV2/cache/schema');
+
+const { generateFakeEvents } = require('../../../utils/v2Data');
+
+const startTime = new Date().getTime();
+const endTime = startTime + 30000; // Add 30 seconds;
+const testValues = generateFakeEvents(startTime, endTime, 50);
+
+describe('Test cache memory backend', () => {
```
jonathan-gramain, line 12: Is there a way to use fixtures to test multiple backends in the same way? As this test code looks similar to the Redis backend.

jonathan-gramain (approved):
Looks good, maybe this PR could have been split into smaller PRs

tests/unit/v2/cache/testMemoryBackend.js
tmacro: I considered it, but decided that given the direct introspection of the db I in the memory backend that it would lead to some sort of wrapper to abstract the asserts that would have to be maintained as well. Since the tested API is small and the tests relatively short I think the added cost of abstraction would outweigh the maintaining separate tests.

tests/functional/v2/cache/testRedisBackend.js
tmacro: Internally shards are just treated as opaque strings used to generate a key so I don't think it's an issue.

libV2/cache/backend/memory.js
tmacro: This is really just to match the redis client where the return value signifies whether the redis command succeeded

libV2/utils/shard.js
tmacro: I've thought of a way to do it with no floats.

tmacro:
@jonathan-gramain It grew a bit in size after I opened it. Ended up a little bigger than planned.

naren-rajendran (approved):

PR #481, opened 2020-06-18
ft(S3C-3004): Add config loading

tmacro:


libV2/config/index.js
```diff
@@ -0,0 +1,223 @@
+const fs = require('fs');
+const path = require('path');
+const Joi = require('@hapi/joi');
+const { truthy } = require('../constants');
+
+const { envNamespace } = require('../constants');
```
miniscruff: Can merge these two
```suggestion
const { truthy, envNamespace } = require('../constants');
```

libV2/config/index.js
```diff
@@ -0,0 +1,223 @@
...
+    bool: val => truthy.includes(val.toLowerCase()),
+    int: val => parseInt(val, 10),
+    list: val => val.split(',').map(v => v.trim()),
+};
+
+function _definedInEnv(key) {
+    return process.env[`${envNamespace}_${key}`] !== undefined;
+}
+
+function _loadFromEnv(key, defaultValue, type) {
```
miniscruff, line 19: Instead of the type being a string of that type, what if it was the typecast method. So instead of passing in 'bool', you would pass in `_typeCasts.bool`. I prefer not to use strings when an object or literally anything else also works.

libV2/config/index.js
```diff
@@ -0,0 +1,223 @@
...
+    if (_definedInEnv(key)) {
+        const envKey = `${envNamespace}_${key}`;
+        const value = process.env[envKey];
```
miniscruff: these two lines are copied from _definedInEnv, not sure the best way to merge the two but I am sure one exists.

libV2/config/index.js
```diff
@@ -0,0 +1,223 @@
...
+    static _readFile(path) {
+        try {
+            const data = fs.readFileSync(path, {
+                encoding: 'utf-8',
+            });
+            return JSON.parse(data);
+        } catch (error) {
+            // eslint-disable-next-line no-console
+            console.error(`error reading config file at ${path}`, { error });
+            throw new Error(`could not parse config file: ${error.message}`);
```
miniscruff: I would remove the try-catch as it does not add much value to the existing errors that readFileSync and JSON.parse will already throw on bad issues and just adds more work.

libV2/constants.js
```diff
@@ -55,4 +55,13 @@ module.exports = {
...
     warp10ValueType: ':utapi:event',
+    truthy: [
+        'true',
+        'on',
+        'yes',
+        'y',
+        't',
+        'enabled',
+        'enable',
+    ],
```
alexanderchan-scality: a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) can be used rather than an array.

libV2/config/index.js
```diff
@@ -0,0 +1,223 @@
...
+            .concat(Object.keys(child))
+            .reduce((keys, key) => {
+                if (!keys.includes(key)) keys.push(key);
+                return keys;
+            }, []);
+        return keys.reduce((ret, key) => {
+            // eslint-disable-next-line no-param-reassign
+            ret[key] = this._recursiveUpdate(parent[key], child[key]);
+            return ret;
+        }, {});
```
miniscruff: I mean you have gone this far might as well do it all in one statement :)

```suggestion
        return Object.keys(parent)
            .concat(Object.keys(child))
            .reduce((keys, key) => {
                if (!keys.includes(key)) {
                    keys.push(key);
                }
                return keys;
            }, [])
            .reduce((current, key) => Object.assign(
                current, 
                {[key]: this._recursiveUpdate(parent[key], child[key]) }
            ), {});
```

miniscruff (changes requested):

libV2/constants.js
tmacro: What upsides does a set have here over a static array?

miniscruff: Sets are generally better for pure lookups as they are hashed so O(1) lookups.

miniscruff: Preferably if javascript would freeze them that would be cool but there is no immutable sets in the standard packages as far as I know. Python has frozen sets https://docs.python.org/3/library/stdtypes.html#frozenset which are even better.

libV2/config/index.js
tmacro: I added it here to make sure the something is written to the console in case of error as the config module is wired in pretty deeply and I don't want to accidentally silently eat the error

miniscruff: Hmm, I guess that is fine but it will be pretty hard to silently eat all the errors that will come from this I think.
You could also just rethrow the error and keep the error log.

libV2/constants.js
tmacro: makes sense

libV2/config/index.js
miniscruff: @tmacro mind letting me resolve comments as you update them? It is how I know whether or not to approve by going through the unresolved comments. And when they get resolved I get confused lol.

libV2/config/index.js
```diff
@@ -0,0 +1,222 @@
...
+        this.healthChecks = undefined;
+        this.logging = { level: 'debug', dumpLevel: 'error' };
+
+        this.redis = undefined;
+        this.warp10 = undefined;
+
+        // read config automatically
+        const loadedConfig = this._loadConfig();
+        let parsedConfig = Config._parseConfig(loadedConfig);
+        if (overrides && typeof overrides === 'object') {
```
jonathan-gramain: I think `typeof overrides === 'object'` implies a truish `overrides` so the first check must be superfluous

libV2/config/index.js
```diff
@@ -0,0 +1,222 @@
...
+        const ret = [];
+        for (let i = 0; i < Math.max(parent.length, child.length); i += 1) {
+            ret[i] = this._recursiveUpdate(parent[i], child[i]);
+        }
+        return ret;
+    }
+
+    _recursiveUpdateObject(parent, child) {
+        return Object.keys(parent)
+            .concat(Object.keys(child))
```
jonathan-gramain: You can probably replace the `concat` and first `reduce` by constructing a `Set` of keys.

libV2/config/index.js
```diff
@@ -0,0 +1,222 @@
...
+        if (Array.isArray(parent) && Array.isArray(child)) {
+            return this._recursiveUpdateArray(parent, child);
+        }
+        if (typeof parent === 'object' && typeof child === 'object') {
+            return this._recursiveUpdateObject(parent, child);
+        }
+        return child;
+    }
+
+    _loadConfig() {
```
jonathan-gramain, line 136: I wonder if it would not be better to use `joi` directly to map the user config over the default config. In my memories, joi allows to set defaults for any value or entire object, wondering if it would work for what you want to achieve.

libV2/constants.js
```diff
@@ -1,3 +1,14 @@
+
+const truthy = new Set([
```
jonathan-gramain, line 2: if you want to be exhaustive, `1` is also a commonly used value for a boolean.

libV2/constants.js
```diff
@@ -1,3 +1,14 @@
...
+    'true',
+    'on',
+    'yes',
+    'y',
```
jonathan-gramain, line 6: I've never seen `y` and `t` used in practice in our product, but why not

jonathan-gramain (approved):
Only minor comments, LGTM otherwise

libV2/config/index.js
tmacro: sure, I was resolving so I could keep track :)

libV2/config/index.js
tmacro: I tried this out in jabba and found that with some of the nesting it can get pretty hairy handling the defaults. Plus I've been using some form of the recursiveUpdate logic in projects for years so its pretty battle tested

libV2/constants.js
miniscruff: the python stdlib has a method similar to this https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool

tmacro: lol thats where the list originally came from

miniscruff (approved):

PR #492, opened 2020-06-18
ft(S3C-3004): Add process abstraction

tmacro:


libV2/process.js
```diff
@@ -0,0 +1,31 @@
...
+        await this._setup();
+    }
+
+    async start() {
+        await this._start();
+    }
+
+    async join() {
+        this.emit('exit');
+        await this._join();
```
alexanderchan-scality, line 26: needs clarification: where are these private methods coming from?

tmacro: I'll stub them out for clarity, these would be overridden by subclasses

alexanderchan-scality: should check for existence first before calling since there is no guarantee they will be overwritten? or create a noop method?

tmacro: I'll do a noop on the base `Proccess` class so they always exist

libV2/process.js
```diff
@@ -0,0 +1,40 @@
...
+const { logger } = require('./utils');
+
+
+class Process extends EventEmitter {
+    async setup() {
+        const cleanUpFunc = this.join.bind(this);
+        process.on('SIGINT', cleanUpFunc);
+        process.on('SIGHUP', cleanUpFunc);
+        process.on('SIGQUIT', cleanUpFunc);
+        process.on('SIGTERM', cleanUpFunc);
```
miniscruff: A loop?
```suggestion
        ['SIGINT', 'SIGHUP', 'SIGQUIT', 'SIGTERM'].forEach(eventName => {
            process.on(eventName, cleanUpFunc);
        };
```

Although, I am not sure if this will mess up what `this` refers to inside the cleanup function. So this may or may not work...

miniscruff (approved):

libV2/process.js
```diff
@@ -0,0 +1,38 @@
...
+class Process extends EventEmitter {
+    async setup() {
+        const cleanUpFunc = this.join.bind(this);
+        ['SIGINT', 'SIGHUP', 'SIGQUIT', 'SIGTERM'].forEach(eventName => {
+            process.on(eventName, cleanUpFunc);
+        };
+        process.on('uncaughtException', error => {
+            logger.error('uncaught exception',
+                { error, stack: error.stack.split(os.EOL) });
+            return cleanUpFunc();
```
miniscruff: Should we return this?

libV2/process.js
```diff
@@ -0,0 +1,38 @@
+const { EventEmitter } = require('events');
+const os = require('os');
+const { logger } = require('./utils');
+
+
+class Process extends EventEmitter {
+    async setup() {
+        const cleanUpFunc = this.join.bind(this);
+        ['SIGINT', 'SIGHUP', 'SIGQUIT', 'SIGTERM'].forEach(eventName => {
```
jonathan-gramain: not sure `SIGHUP` is a good candidate for termination, it's usually when the terminal attached to the process is closed, which would not make much sense if the process is a daemon.

alexanderchan-scality (approved):

PR #500, opened 2020-06-22
Feature/s3 c 3004 add http server

tmacro:


ilkescality (approved):
LGTM (nice to see spread operators asyncs awaits)

libV2/cache/index.js
```diff
@@ -1,7 +1,27 @@
...
+    backend = new MemoryCache();
+    break;
+
+case 'redis':
+    backend = new RedisCache(config.redis);
+    break;
+
+default:
+    throw Error(`Invalid cache backend ${config.cacheBackend}`);
+}
```
miniscruff: What about using a map/object to go from cache backend string to cache object? Or instead of using cache backend just the presence of 'memory'/'redis' would use that object? Maybe something like:

```suggestion
const cacheTypes = {
  redis: RedisCache,
  memory: MemoryCache,
};
const backend = new cacheTypes[config.cacheBackend](config[config.cacheBackend] || {});
// this will throw an error if cache backend is undefined we may want to handle that...
```

libV2/errors/errors.json
```diff
@@ -0,0 +1,30 @@
...
+    },
+    "OperationTimedOut": {
+        "code": 500,
+        "description": "The operation could not be completed within the permitted time."
+    },
+    "ServerBusy": {
+        "code": 503,
+        "description": "The server (or an internal component) is currently unavailable to receive requests. Please retry your request."
+    }
+}
```
miniscruff: should add a newline here

miniscruff (changes requested):
Needs more tests for sure but looks good.
Doing request changes just so my comments can be resolved first.

libV2/errors/index.js
```diff
@@ -0,0 +1,27 @@
...
+function errorsGen() {
+    return Object.keys(utapiErrors)
+        .reduce((errors, name) => {
+            errors[name] = new UtapiError(name, utapiErrors[name].code,
+                utapiErrors[name].description);
+            return errors;
+        }, {});
+}
+
+module.exports = errorsGen();
```
ilkescality, line 27: noticed parens here, not sure if intentional

libV2/cache/index.js
tmacro: I tried a variation of your suggestion that I thinks a bit more readable.

libV2/errors/index.js
tmacro: It's intentional, it exports a map of instantiated Errors

miniscruff (approved):

libV2/cache/index.js
miniscruff: Agreed

tests/unit/v2/testMiddleware.js
```diff
@@ -18,4 +19,60 @@ describe('Test middleware', () => {
...
+            middleware.errorMiddleware({ code: 123, message: 'Hello World!', utapiError: true }, null, resp);
+            assert.deepStrictEqual(resp._body, {
+                error: {
+                    code: '123',
+                    message: 'Hello World!',
+                },
+            });
+        });
+
+        it('should replace a error\'s message if it\'s internal and not in development mode', () => {
```
miniscruff: could use double quotes to avoid escaping single quotes, I think eslint lets you do that

tests/unit/v2/models/testResponseContainer.js
```diff
@@ -0,0 +1,85 @@
...
+        });
+
+        it('should not allow status codes < 100', () => {
+            let passed = false;
+            try {
+                resp.statusCode = 99;
+            } catch (err) {
+                passed = true;
+            }
+            assert(passed);
```
miniscruff: Can use assert.throws here
```suggestion
           assert.throws(() => res.statusCode = 99);
```

tests/unit/v2/models/testResponseContainer.js
```diff
@@ -0,0 +1,85 @@
...
+            try {
+                resp.redirect = '\\foo\\bar';
+            } catch (err) {
+                passed = true;
+            }
+            assert(passed);
+        });
+
+        it('should not allow a non-http scheme', () => {
+            let passed = false;
```
miniscruff: assert.throws here as well

tests/unit/v2/models/testResponseContainer.js
```diff
@@ -0,0 +1,85 @@
...
+        it('should not have a redirect', () => {
+            assert.strictEqual(resp.hasRedirect(), false);
+        });
+
+        it('should allow a relative redirect', () => {
+            resp.redirect = 'im/relative';
+        });
+
+        it('should not allow invalid redirect', () => {
+            let passed = false;
```
miniscruff: assert.throws here as well

miniscruff (changes requested):
Just some assert throws usages I think would look good, getting kinda big though might want to get an extra reviewer or two

miniscruff (approved):

miniscruff:
@tmacro would you mind using the WIP label or something else to signify you are still actively making changes and that its not quite ready. Or something similar anyway.

libV2/server/controller.js
```diff
@@ -0,0 +1,157 @@
...
+            .then(
+                done,
+                done,
+            );
+    }
+
+    /**
+     * buildMap
+     *
+     * Constructs a object of `operationId`|`callOperation` pairs for use as a controller with oas-tools
```
miniscruff:
```suggestion
     * Constructs an object of `operationId`|`callOperation` pairs for use as a controller with oas-tools
```

libV2/server/controller.js
```diff
@@ -0,0 +1,157 @@
...
+     * @returns {undefined}
```
miniscruff: I don't think this returns undefined

tests/utils/v2Data.js
```diff
@@ -65,6 +67,75 @@ function generateFakeEvents(start, stop, count) {
...
+        return this;
+    }
+
+    redirect(url) {
+        this._redirect = url;
+        return this;
+    }
+}
+
+const stubLogger = {
```
miniscruff, line 98: Maybe use sinon to create the stubs?

tests/unit/v2/server/testMiddleware.js
```diff
@@ -0,0 +1,78 @@
...
+            middleware.errorMiddleware({ code: 123, message: 'Hello World!', utapiError: true }, null, resp);
+            assert.deepStrictEqual(resp._body, {
+                error: {
+                    code: '123',
+                    message: 'Hello World!',
+                },
+            });
+        });
+
+        it("should replace a error's message if it's internal and not in development mode", () => {
```
miniscruff:
```suggestion
        it("should replace an error's message if it's internal and not in development mode", () => {
```

tests/unit/v2/server/testController.js
```diff
@@ -0,0 +1,113 @@
...
+    });
+
+    it('should load an operation handler', () => {
+        const handler = APIController._getOperationHandler('internal', 'healthcheck');
+        assert(typeof handler === 'function');
+    });
+
+    it('should patch a NotImplemented handler for MODULE_NOT_FOUND errors', () => {
+        const handler = APIController._getOperationHandler('invalid', 'idontexist');
+        assert(typeof handler === 'function');
```
miniscruff, line 33: just checking it returns a method doesn't really confirm much imo, maybe assert the returned function returns something or it matches when calling the internal method directly?

tests/unit/v2/server/testController.js
```diff
@@ -0,0 +1,113 @@
...
+        const req = {
+            swagger: {
+                params: {
+                    foo: {
+                        value: 'bar',
+                    },
+                },
+            },
+        };
+        assert(APIController._extractParams(req), { foo: 'bar' });
```
miniscruff: did you mean to assert deep equals here? This just checks _extractParams returns a truthy value I believe

tests/unit/v2/models/testRequestContext.js
```diff
@@ -0,0 +1,124 @@
...
+            request: testCase.request,
+            logger: testCase.request.logger,
+            results: testCase.request.results,
+            ...(testCase.expected || {}),
+        });
+
+        try {
+            ctx = new RequestContext(testCase.request);
+        } catch (err) {
+            if (testCase.valid) {
```
miniscruff: If you split the valid tests from invalid tests this becomes a little easier as you wouldn't need a few of the branches. Try/catch won't be required for the valid cases and we can assert throws the invalid cases.

miniscruff (changes requested):

libV2/server/controller.js
```diff
@@ -0,0 +1,157 @@
...
+            }
+            return null;
+        }
+    }
+
+    static _notImplementedHandler(tag, operationId) {
+        // eslint-disable-next-line no-unused-vars
+        return async (ctx, params) => {
+            throw errors.NotImplemented.customizeDescription(
+                `the operation "${tag}::${operationId}"  has not been implemented`,
```
miniscruff:
```suggestion
                `the operation "${tag}::${operationId}" has not been implemented`,
```

tests/unit/v2/server/testMiddleware.js
miniscruff: If you do decide to just take my suggestion as is you can then resolve the comment, or maybe github auto resolves it idk. I kinda wish github would say if the suggestion was used or not...

tests/utils/v2Data.js
tmacro: I'd like to leave this rather dependency agnostic so tests not otherwise using sinon don't have to worry about side effects. Additionally stubbing it with sinon wouldn't net us much in the way of brevity or simplicity.

tests/unit/v2/server/testController.js
```diff
@@ -0,0 +1,120 @@
...
+
+    it('should load an operation handler', () => {
+        const handler = APIController._getOperationHandler('internal', 'healthcheck');
+        assert(typeof handler === 'function');
+    });
+
+    it('should patch a NotImplemented handler for MODULE_NOT_FOUND errors', async () => {
+        const handler = APIController._getOperationHandler('invalid', 'idontexist');
+        assert(typeof handler === 'function');
+        assert.rejects(handler);
```
miniscruff: assert rejects allows you to pass an error validation method in instead of calling the handler a second time inside a try catch

tests/unit/v2/models/testRequestContext.js
```diff
@@ -0,0 +1,112 @@
...
+                assert.deepStrictEqual(ctx.getValue(), expected);
+            }));
+    });
+
+    describe('Test invalid cases', () => {
+        invalidTestCases.forEach(testCase =>
+            it('should fail to create a RequestContext', () => {
+                assert.throws(() => new RequestContext(testCase.request));
+            }));
+    });
```
miniscruff, line 111: :sparkles: It's so clean :sparkles:

miniscruff (approved):

PR #501, opened 2020-06-26
Feature/s3 c 3004 debugging

tcarmet:


eve/main.yml
```diff
@@ -30,11 +33,7 @@ stages:
     steps:
     - MasterShellCommand:
         name: Replace upstream image with `-ci` variant
-        command: "sed '/^FROM/ s/$/-ci/' images/warp10/Dockerfile"
-        workdir: "%(prop:master_builddir)s/build/"
```
tcarmet, line 34: don't know why but the `workdir` parameter is not available on the `MasterShellCommand` step, therefore I've added the property directly on the command itself

eve/main.yml
```diff
@@ -17,6 +17,9 @@ models:
       path: eve/workers/pod.yml
       images:
         aggressor: eve/workers/unit_and_feature_tests
+        warp10:
```
tcarmet, line 20: I'm letting eve build the warp10 image as it needs to be published on a registry in order that pods in other nodes can pick up the image.

eve/main.yml
```diff
@@ -30,11 +33,7 @@ stages:
     steps:
     - MasterShellCommand:
         name: Replace upstream image with `-ci` variant
-        command: "sed '/^FROM/ s/$/-ci/' images/warp10/Dockerfile"
-        workdir: "%(prop:master_builddir)s/build/"
-    - ShellCommand:
-        command: docker build -f images/warp10/Dockerfile -t warp10 .
```
tcarmet, line 36: This is now done by eve

PR #506, opened 2020-06-29
ft(S3C-3008): Add Utapi client

tmacro:


libV2/client/index.js
```diff
@@ -0,0 +1,248 @@
...
+            await this._pushToUtapi([metric]);
+        } catch (error) {
+            this._logger.error('unable to push metric, adding to retry cache', { error });
+            if (!await this._addToRetryCache(metric)) {
+                throw new Error('unable to store metric');
+            }
+        }
+    }
+
+    pushMetric(data, cb) {
```
miniscruff, line 255: I am not sure I understand why cb is here. It seems to be passed in to the _pushMetric method but that doesnt accept a callback.

tmacro: This is to keep compatibility with the existing Utapi client which accepts a callback. If callback is provided `callbackify` is used to convert the promise return by `_pushMetric` to callback style otherwise a Promise is returned

miniscruff: ok that makes sense then. Is that the same reason for the other callbacks sprinkled around?

tmacro: The callbacks in `Chunker` and `Uploader` are inherited. The test use whatever I thought would be the most simple, though I did try to lean more toward testing `pushMetric` with callbacks as that's what cloudserver would currently be using

miniscruff: Gotcha I see

miniscruff (approved):

libV2/client/index.js
```diff
@@ -0,0 +1,248 @@
...
+                callback();
+            });
+    }
+}
+
+class UtapiClient {
+    constructor(config) {
+        this._host = (config && config.host) || 'localhost';
+        this._port = (config && config.port) || '8100';
+        this._logger = (config && config.port) || moduleLogger;
```
jonathan-gramain: I guess it should be `config && config.logger`

libV2/client/index.js
```diff
@@ -0,0 +1,248 @@
...
+const moduleLogger = new LoggerContext({
+    module: 'client',
+});
+
+const _noOp = () => {};
+
+class Chunker extends Transform {
+    constructor(options) {
+        super({ objectMode: true, ...options });
+        this._chunkSize = options.chunksize || 100;
```
jonathan-gramain: minor point: could use more standard camel case for `chunksize`: `chunkSize`

libV2/client/index.js
```diff
@@ -0,0 +1,248 @@
...
+
+    async _addToRetryCache(metric) {
+        if (this._numCachedMetrics < this._maxCachedMetrics) {
+            try {
+                await this._retryCache.put(metric.uuid, metric.getValue());
+                this._numCachedMetrics += 1;
+                await this._scheduleDrain();
+                return true;
+            } catch (error) {
+                this._logger
```
jonathan-gramain, line 108: might be useful to log the `metric` object as well if not too big, although it's logged in the recovery log, it could maybe help to correlate this error with the recovery log.

libV2/client/index.js
```diff
@@ -0,0 +1,248 @@
...
+                .pipe(new Uploader({ ingest: this._pushToUtapi.bind(this) }))
+                .on('data', res => {
+                    if (res.success) {
+                        toRemove.push(...res.keys);
+                    } else {
+                        success = false;
+                    }
+                })
+                .on('end', () => {
+                    this._retryCache.batch(toRemove.map(i => ({ type: 'del', key: i })));
```
jonathan-gramain: The `i` variable is generally used for iteration index, might be better to use an alternate name to avoid confusion like simply `key` or `k`

libV2/client/index.js
```diff
@@ -0,0 +1,248 @@
...
+        const toRemove = [];
+
+        return new Promise((resolve, reject) => {
+            this._retryCache.createReadStream()
+                .on('data', entry => {
+                    this._emitMetricLogLine(entry.value);
+                    toRemove.push(entry.key);
+                })
+                .on('end', () => {
+                    this._retryCache.batch(toRemove.map(i => ({ type: 'del', key: i })));
```
jonathan-gramain: same about using the variable `i`

openapi.yaml
```diff
@@ -5,6 +5,38 @@ info:
...
+          location:
+            type: string
+          objectDelta:
+            type: integer
+          sizeDelta:
+            type: integer
+          incomingBytes:
+            type: integer
+          outgoingBytes:
+              type: integer
```
jonathan-gramain: indentation seems too large here

tests/unit/v2/testClient.js
```diff
@@ -0,0 +1,143 @@
...
+
+    afterEach(() => {
+        sandbox.restore();
+        client = null;
+    });
+
+    it('should push a metric', next => {
+        sandbox.stub(UtapiClient.prototype, '_pushToUtapi').resolves();
+        client.pushMetric(events[0], err => {
+            assert.strictEqual(err, null);
```
jonathan-gramain: you may use `assert.ifError(err)`, it would also accept `undefined` and might have a better error reporting

tests/unit/v2/testClient.js
```diff
@@ -0,0 +1,143 @@
...
+            next();
+        });
+    });
+
+    it('should add a metric to the retry cache if an error occurs during pushing', next => {
```
jonathan-gramain: Since each test is supposedly independent and not part of a series, `next` could be renamed to `done` or `cb`

tests/unit/v2/testClient.js
```diff
@@ -0,0 +1,143 @@
...
+        sandbox.stub(UtapiClient.prototype, '_pushToUtapi').rejects();
+        const spy = sandbox.spy(client, '_addToRetryCache');
+        client.pushMetric(events[0], err => {
+            assert.strictEqual(err, null);
+            assert(spy.calledWith(events[0]));
+            assert(client._drainTimer !== undefined);
```
jonathan-gramain: could also be `assert.notStrictEqual(client._drainTimer, undefined)`

libV2/client/index.js
```diff
@@ -0,0 +1,248 @@
...
+            return resp.statusCode === 200;
+        } catch (error) {
+            this._logger
+                .debug('drain preflight request failed', { error });
+            return false;
+        }
+    }
+
+    async _attemptDrain() {
+        this._drainTimer = null;
```
jonathan-gramain: It could be better to reset the `drainTimer` only after the drain is complete to avoid possibly a new drain to kick in while this one is still in progress

libV2/client/index.js
```diff
@@ -0,0 +1,248 @@
...
+            }
+        } else {
+            this._emitMetricLogLine(metric, { reason: 'overflow' });
+        }
+        return false;
+    }
+
+    async _drainRetryCache() {
+        return new Promise((resolve, reject) => {
+            let success = true;
```
jonathan-gramain: I find it a bit confusing that this variable is called `success` while the caller in `_attemptDrain` calls it `empty`, maybe should have the same name

libV2/client/index.js
jonathan-gramain: This call does not have an `await`, is it on purpose, like not to wait for this call to complete?

libV2/client/index.js
tmacro: we can log the uuid of the metric and avoid logging the whole thing

jonathan-gramain (approved):

PR #517, opened 2020-07-01
ft(S3C-3005): Ingest API

tmacro:


alexanderchan-scality (approved):

libV2/server/middleware.js
```diff
@@ -39,9 +40,18 @@ function loggerMiddleware(req, res, next) {
 // next is purposely not called as all error responses are handled here
 // eslint-disable-next-line no-unused-vars
 function errorMiddleware(err, req, res, next) {
-    const code = err.code || 500;
+    let code = err.code || 500;
     let message = err.message || 'Internal Error';
 
+    // failed request validation by oas-tools
+    if (err.failedValidation) {
+        // You can't actually use destructing here
```
miniscruff, line 48: why not?

tmacro:
```
Tmac@Kennuc: /tmp cat test.js
let x;
let y;

{ x, y } = { x:1, y:2}
Tmac@Kennuc: /tmp node test.js
/tmp/test.js:4
{ x, y } = { x:1, y:2}
         ^

SyntaxError: Unexpected token '='
    at wrapSafe (internal/modules/cjs/loader.js:1054:16)
    at Module._compile (internal/modules/cjs/loader.js:1102:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1158:10)
    at Module.load (internal/modules/cjs/loader.js:986:32)
    at Function.Module._load (internal/modules/cjs/loader.js:879:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at internal/main/run_main_module.js:17:47
```

miniscruff: hmm seems like it doesn't like destructuring onto existing objects... good to know. Not sure if the comment is required but yea.

miniscruff: Note that from the node REPL it works just fine which is confusing...
```shell
Welcome to Node.js v14.1.0.
Type ".help" for more information.
> let x, y;
undefined
> { x, y } = { x:1, y:2 }
{ x: 1, y: 2 }
> console.log(x)
1
undefined
> console.log(y)
2
undefined
```

tests/unit/v2/API/metrics/testIngestMetric.js
```diff
@@ -0,0 +1,38 @@
...
+
+const events = generateFakeEvents(1, 50, 50);
+
+describe('Test ingestMetric', () => {
+    let ctx;
+    beforeEach(() => {
+        ctx = templateContext();
+    });
+
+    afterEach(() => sinon.restore());
```
miniscruff, line 16: I think this also works
```suggestion
    afterEach(sinon.restore);
```

libV2/server/API/metrics/ingestMetric.js
```diff
@@ -0,0 +1,20 @@
...
+    try {
+        metrics = params.body.map(m => new UtapiMetric(m));
+    } catch (error) {
+        throw errors.InvalidRequest;
+    }
+    try {
+        await Promise.all(metrics.map(m => cacheClient.pushMetric(m)));
+    } catch (error) {
+        throw errors.ServiceUnavailable;
+    }
```
miniscruff, line 16: You could combine the catches I think
```suggestion
    try {
        let metrics = params.body.map(m => new UtapiMetric(m));
        await Promise.all(metrics.map(m => cacheClient.pushMetric(m)));
    } catch (error) {
        throw errors.InvalidRequest || errors.ServiceUnavailable;
    }
```

Not sure if this any more readable but maybe.

miniscruff (approved):
Small ideas looks good otherwise

libV2/server/API/metrics/ingestMetric.js
tmacro: Wouldn't this `throw errors.InvalidRequest || errors.ServiceUnavailable;` always throw `errors.InvalidRequest`?

miniscruff: Edit: I misread the original code oops... I thought it was `error.InvalidRequest` as in the error was raised which had an object as part of it. And not just grabbing an error from the errors object.

tests/unit/v2/API/metrics/testIngestMetric.js
tmacro: sinon looses some sort of internal state without the `() =>` and throws an error

miniscruff: hmm weird, will have to remember this for later

PR #526, opened 2020-07-13
Feature/s3 c 3001 add redis to warp10 task

tmacro:


images/warp10/Dockerfile
```diff
@@ -3,16 +3,22 @@ FROM warp10io/warp10:2.6.0
...
+# ENV warpscript.extension.debug io.warp10.script.ext.debug.DebugWarpScriptExtension
+
 RUN wget https://github.com/just-containers/s6-overlay/releases/download/v${S6_VERSION}/s6-overlay-amd64.tar.gz -O /tmp/s6-overlay-amd64.tar.gz \
     && tar xzf /tmp/s6-overlay-amd64.tar.gz -C / \
     && rm -rf /tmp/s6-overlay-amd64.tar.gz
 
-# Modify Warp 10 default config
-RUN sed -i "s~^warpscript.repository.directory = .*~warpscript.repository.directory = /usr/local/share/warpscript~" ${WARP10_CONF_TEMPLATES}/* \
-    && sed -i -e 's|^standalone\.host.*|standalone.host = 0.0.0.0|g' ${WARP10_CONF_TEMPLATES}/*
+# Install  protobuf extestion
```
jonathan-gramain: ~extestion~ extension + double spacing `Install  protobuf`

images/warp10/s6/cont-init.d/40-generate-secrets
```diff
@@ -10,10 +10,10 @@ WARP10_CONFIG_DIR="$WARP10_DATA_DIR/conf"
 INITIAL_TOKENS="$WARP10_CONFIG_DIR/initial.token"
 
 if [ ! -f "$INITIAL_TOKENS" ]; then
-    CONFIG_FILES="$(find ${WARP10_CONFIG_DIR} -not -path "*/\.*" -name "*.conf" | sort | tr '\n' ' ' 2> /dev/null)"
+    CONFIG_FILES="$(find ${d} -not -path "*/\.*" -name "*.conf" | sort | tr '\n' ' ' 2> /dev/null)"
```
jonathan-gramain: What is `${d}` here?

images/warp10/s6/cont-init.d/40-generate-secrets
```diff
@@ -10,10 +10,10 @@ WARP10_CONFIG_DIR="$WARP10_DATA_DIR/conf"
...
 
     # Look for a set token secret and use it for generation
-    secret=`${JAVA} -cp ${WARP10_CP} io.warp10.WarpConfig ${CONFIG_FILES} 'token.secret' | grep 'token.secret' | sed -e 's/^.*=//'`
+    secret=`su ${WARP10_USER} -c "${JAVACMD} -cp ${WARP10_CP} io.warp10.WarpConfig ${CONFIG_FILES} . 'token.secret' | grep -e '^@CONF@ ' | sed -e 's/^@CONF@ //' | grep 'token.secret' | sed -e 's/^.*=//'"`
```
jonathan-gramain: Each sed / grep couple can be combined in a single sed with `-n` option (no printing of lines by default) and `s///p` to print sed-ed lines, like:

```
sed -n 's/^@CONF@ //p' | sed -n 's/^token.secret[^=]*=//p'
```

Not sure the syntax is right and not sure about the second command because it depends on the format of the lines, but you get the idea.

libV2/config/index.js
```diff
@@ -206,6 +208,10 @@ class Config {
 
         parsedConfig.cacheBackend = _loadFromEnv('CACHE_BACKEND', config.cacheBackend);
 
+        parsedConfig.ingestionSchedule = _loadFromEnv('INGESTION_SCHED', config.ingestionSchedule);
```
jonathan-gramain: why not `INGESTION_SCHEDULE` to avoid confusion (not worth saving 3 chars IMHO), same for the others.

Generally speaking, it could be a good idea to prefix all env with `UTAPI_` for proper namespacing, but might be too late now.

libV2/process.js
```diff
@@ -1,9 +1,14 @@
 const { EventEmitter } = require('events');
 const os = require('os');
+const { Command } = require('commander');
 const { logger } = require('./utils');
 
-
 class Process extends EventEmitter {
+    constructor(...options) {
+        super(...options);
+        this._program = new Command();
```
jonathan-gramain: Does it need to be created here since it's also done in the `setup` function?

libV2/tasks/BaseTask.js
```diff
@@ -0,0 +1,82 @@
...
+                'Execute task using this crontab.  Overrides configured schedule',
+                value => {
+                    cronparser.parseExpression(value);
+                    return value;
+                },
+            );
+    }
+
+    get schedule() {
+        if (this._program.now) return Now;
```
jonathan-gramain: I'd recommend always using braces around `if`, `for` etc. and put the single instruction on the next line, because it makes the instruction more visible, and could avoid any ambiguity with `if/else` constructs that could be nested. (I think at least in some projects the linter enforces this).

libV2/tasks/IngestShard.js
```diff
@@ -0,0 +1,54 @@
...
+    async _execute(timestamp) {
+        const endShard = shardFromTimestamp(timestamp - (shardIngestLagSecs * 1000000));
+        logger.debug('ingesting shards', { endShard });
+
+        const available = await this._cache.getShards();
+        const toIngest = available
+            .map(key => key.split(':')[2])
+            .filter(key => parseInt(key, 10) <= endShard);
+
+        if (!toIngest.length) {
```
jonathan-gramain: I'd rather avoid using boolean operators on non-booleans (even though some C programmers like to do it :laughing: ), doing an explicit comparison against 0 is more readable.

Note that javascript allows to check for object existence like `if (obj)` or `if (!obj)`, this is a different use that seems more appropriate since an object exists or not (boolean).

libV2/tasks/IngestShard.js
```diff
@@ -0,0 +1,54 @@
...
+            logger.debug('no shard available to ingest');
+            return;
+        }
+
+        await Promise.all(toIngest.map(
+            async shard => {
+                if (await this._cache.shardExists(shard)) {
+                    const metrics = await this._cache.getMetricsForShard(shard);
+                    if (metrics.length) {
```
jonathan-gramain: same here: `if (metrics.length > 0)`

libV2/tasks/IngestShard.js
```diff
@@ -0,0 +1,54 @@
...
+                        logger.info(`Ingesting ${metrics.length} events from shard`, { shard });
+                        const records = metrics
+                            .map(m => new UtapiMetric(JSON.parse(m)));
+                        const status = await this._warp10.ingest('utapi.event', records);
+                        assert.strictEqual(status, records.length);
```
jonathan-gramain, line 41: I'm personally ok with using assertions in the code (some people may disagree) but make sure it's really an assertion and not an error case to handle, so by design `warp10.ingest` must return as many statuses as records, whatever happens. Did not check this but wanted to make the point anyway.

images/warp10/s6/cont-init.d/40-generate-secrets
tmacro: A bug :) It should have not been changed. I've got to check and see why it didn't break the CI as it shouldn't be working if this broken

libV2/config/index.js
tmacro: `_loadFromEnv` prefixes it for you

libV2/tasks/IngestShard.js
tmacro: I was using `assert` as a more compact way to throw an `Error` here, and as plus you get a nicely formatted error message

jonathan-gramain: Ah ok, why not, even though we tend to use assert e.g. for config parsing to fail early, I think it better to explicitly throw errors and reserve asserts for programming logic errors, but up to you I'm good with that too.

jonathan-gramain (approved):

ilkescality (approved):

PR #527, opened 2020-07-13
ft(S3C-3002): Add checkpoint creation

tmacro:


dora-korpar (approved):

warpscript/utapi/decodeCheckpoint.mc2
```diff
@@ -0,0 +1,14 @@
...
+  int64 objD = 2;
+  int64 inB = 3;
+  int64 outB = 4;
+  map<string, int64> ops = 5;
+}
+'>
+PROTOC 'proto' STORE
+<%
+  HEX-> !$proto 'Checkpoint' PB->
+%>
```
jonathan-gramain, line 14: Maybe a simple unit test like you did for the snapshot decode/encode could be good to have, same for the other decode/encode macros

warpscript/utapi/sumField.mc2
```diff
@@ -0,0 +1,46 @@
+{
```
jonathan-gramain: Shouldn't this go in warpscript/util directory?

warpscript/utapi/createCheckpoint.mc2
```diff
@@ -0,0 +1,155 @@
+{
```
jonathan-gramain, line 1: This macro looks much like the snapshot one, is there a way to reuse parts of the same code somehow? I'm a noob in mc2 so just take this as a suggestion.

tmacro: There's definitely some generalization that can be done, but as I'm still a noob as well, I've tried to write it in a straightforward manner to try and salvage what readability I can. When I add the repair process in a later PR, I plan to do some refactoring as the requirements will be a bit more clear.

warpscript/utapi/sumField.mc2
tmacro: It's already moved on my local copy :) I'll backport the change

warpscript/utapi/decodeCheckpoint.mc2
tmacro: Added tests for encode/decodeEvent
encodeCheckpoint is generalized to encodeRecord in the snapshot PR with tests

jonathan-gramain (approved):

PR #554, opened 2020-07-23
ft: S3C-3196 update node

dora-korpar:
Installing node modules throws this error:
`error @yarnpkg/core@2.1.1: The engine "node" is incompatible with this module. Expected version ">=10.19.0"`

Updates node version to comply with dependencies.

### DIFF
--- a/eve/workers/unit_and_feature_tests/Dockerfile
+++ b/eve/workers/unit_and_feature_tests/Dockerfile
@@ -1,12 +1,14 @@
 FROM buildpack-deps:jessie-curl
 
 #
-# Install apt packages needed by backbeat and buildbot_worker
+# Install apt packages needed by utapi and buildbot_worker
 #
 ENV LANG C.UTF-8
+ENV NODE_VERSION 10.22.0
 COPY utapi_packages.list buildbot_worker_packages.list /tmp/
 
-RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - \
+RUN wget https://nodejs.org/dist/v${NODE_VERSION}/node-v${NODE_VERSION}-linux-x64.tar.gz \
+    && tar -xf node-v${NODE_VERSION}-linux-x64.tar.gz --directory /usr/local --strip-components 1 \ 
     && curl -sS http://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - \
     && echo "deb http://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list \
     && apt-get update -qq \
@@ -14,7 +16,8 @@ RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - \
     && pip install pip==9.0.1 \
     && rm -rf /var/lib/apt/lists/* \
     && rm -f /tmp/*packages.list \
-    && rm -f /etc/supervisor/conf.d/*.conf
+    && rm -f /etc/supervisor/conf.d/*.conf \
+    && rm -f node-v${NODE_VERSION}-linux-x64.tar.gz
 
 
 #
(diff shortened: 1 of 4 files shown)


eve/workers/unit_and_feature_tests/Dockerfile
```diff
@@ -6,7 +6,7 @@ FROM buildpack-deps:jessie-curl
 ENV LANG C.UTF-8
 COPY utapi_packages.list buildbot_worker_packages.list /tmp/
 
-RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - \
+RUN curl -sL https://deb.nodesource.com/setup_11.x | bash - \
```
miniscruff: does 11.x install node11? I do not think that version is part of LTS, I think they do every other version. Might want to use 12 or an exact version of 10 that is later. Could totally be wrong here tho.

dora-korpar: Yeah, I'm playing around with it right now. I need v10.19.0 but there's no PPAs for specific minor versions :[

miniscruff: lame, I would expect 10.x to install the latest minor+patch version as well.

dora-korpar: yeah, i was surprised by that too. but it installs 10...16? or 14. either way, super lame

eve/workers/unit_and_feature_tests/Dockerfile
```diff
@@ -1,20 +1,22 @@
...
 
 #
-# Install apt packages needed by backbeat and buildbot_worker
+# Install apt packages needed by utapi and buildbot_worker
 #
 ENV LANG C.UTF-8
 COPY utapi_packages.list buildbot_worker_packages.list /tmp/
 
-RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - \
+RUN wget https://nodejs.org/dist/v10.22.0/node-v10.22.0-linux-x64.tar.gz \
```
miniscruff: What about using an env value for the node version? Feels like something worth parameterizing.

dora-korpar:
@bert-e reset

miniscruff (approved):
:+1:

alexanderchan-scality (approved):

ilkescality (approved):

eve/workers/unit_and_feature_tests/Dockerfile
```diff
@@ -1,20 +1,23 @@
 FROM buildpack-deps:jessie-curl
 
 #
-# Install apt packages needed by backbeat and buildbot_worker
+# Install apt packages needed by utapi and buildbot_worker
 #
 ENV LANG C.UTF-8
+ENV NODE_VERSION 10.22.0
```
anurag4DSB, line 7: Are we moving back to v10.22.0 for Utapi?

Had a discussion with @tmacro in one of his older PRs and he wanted to use node 12.
https://github.com/scality/utapi/pull/441#discussion_r430591701 which was targeted for `development/7.7`

anurag4DSB (changes requested):
Adding review as request changes as I am not sure about node version.

anurag4DSB (approved):

eve/workers/unit_and_feature_tests/Dockerfile
dora-korpar: Discussed and resolved :)

tmacro (approved):

eve/workers/unit_and_feature_tests/Dockerfile
tmacro: For some additional context, Utapi v2 is being written be compatible with 12.x with no changes (in fact its what I use to develop) but given the time constraints I decided that it would take to long to update v1 so I left the deployed version as 10.x

dora-korpar:
@bert-e approve

PR #558, opened 2020-07-24
ft(S3C-3007): Add listMetrics handler 

tmacro:


libV2/server/API/metrics/listMetrics.js
```diff
@@ -0,0 +1,82 @@
...
+                    ...operations,
+                },
+            };
+            metric[metricResponseKeys[params.resource]] = result.resource;
+            return metric;
+        });
+
+    // Unwrap the response if only one resource is passed
+    if (resp.length === 1) {
+        [ctx.results.body] = resp;
```
miniscruff, line 74: wait, what does this do? Is this the same as `ctx.results.body = resp[0]`?

tmacro: yeah

PR #565, opened 2020-07-28
ft(S3C-3007): add listMetrics handler

tmacro:


libV2/server/API/metrics/listMetrics.js
```diff
@@ -0,0 +1,82 @@
...
+    const labelName = serviceToWarp10Label[params.resource];
+    const resources = params.body[params.resource];
+    const [start, end] = params.body.timeRange
+        .map(convertTimestamp)
+        .map(v => v.toString());
+
+    // A separate request will be made to warp 10 per requested resource
+    const results = await Promise.all(
+        resources.map(async resource => {
+            const labels = {};
```
jonathan-gramain: A more concise syntax is also possible:
```
const labels = { [labelName]: resource };
```

libV2/server/API/metrics/listMetrics.js
```diff
@@ -0,0 +1,82 @@
...
+                operations: {
+                    ...emptyOperationsResponse,
+                    ...operations,
+                },
+            };
+            metric[metricResponseKeys[params.resource]] = result.resource;
+            return metric;
+        });
+
+    // Unwrap the response if only one resource is passed
```
jonathan-gramain, line 71: Is there one returned resource if and only if only one resource is requested? If so, that looks fine as long as the contract with the API usage is clear, otherwise better keep a consistent API and always return the same data type (array of results, one item per resource?)

By the way, if the API takes an array of resources to query and there's only one resource in that array, I'd rather return an array of one result than a single unwrapped result, also for consistency.

Not sure if this all makes sense, feel free to ignore :smile:

libV2/utils/timestamp.js
```diff
@@ -0,0 +1,18 @@
+/**
+ * Converts the given UNIX timestamp to microsecond precision
+ * @param {Number} timestamp - timestamp to pad
+ * @returns {Number} - Padded timestamp
+ */
+function convertTimestamp(timestamp) {
+    if (timestamp < 10000000000) { // Second precision
```
jonathan-gramain, line 7: This way of doing things looks like a hack to me, even though in practice it should work, does not look very clean. Is there a way to split this function in two different helpers for seconds/milliseconds cases, and call the appropriate helper each time? Or do the math directly instead of calling a helper since it looks like it's a simple multiplication that may not require a helper function.

tests/functional/v2/server/testListMetrics.js
```diff
@@ -0,0 +1,150 @@
...
+            };
+
+            Object.entries(metricQueries)
+                .forEach(query => {
+                    const [level, resources] = query;
+                    it(`should get metrics for ${level}`, async () => {
+                        const resp = await listMetrics(...query, getTs(-500), getTs(0));
+
+                        if (resources.length === 1) {
+                            // If only one resource is requested the response is an object
```
jonathan-gramain, line 123: cf. my earlier comment about API consistency (always returning an array might be better IMHO unless the caller is able to pass an object for a single request for example, in which case we can return a single object since the contract is unambiguous).

As an example of where it can be problematic: if the API caller takes its request array programmatically, it might be annoyed to have an object response instead of an array in case the computed array only has a single element in it.

tests/utils/v2Data/events.js
```diff
@@ -0,0 +1,310 @@
...
+            return params.exists ? 0 : 1;
+        },
+    },
+    delete: {
+        ...objOpFields,
+        sizeDelta: params => {
+            if (params.versioning) {
+                return 0;
+            }
+            return Math.floor((params.currentSize / params.versions)) * -1;
```
jonathan-gramain: minor readability hint: writing `-1 * Math.floor(...)` to emphasize the multiplicator

tests/utils/v2Data/request.js
```diff
@@ -0,0 +1,77 @@
...
+        this._body = null;
+        this._redirect = null;
+    }
+
+    status(code) {
+        this._status = code;
+        return this;
+    }
+
+    sendStatus(code) {
```
jonathan-gramain, line 15: Is there a need for `sendStatus()` when `status()` does the same?

libV2/server/API/metrics/listMetrics.js
tmacro: I agree that a consistent return type is better, but we need to maintain API compatibility with v1.

tests/utils/v2Data/request.js
tmacro: This class is meant to mock the Express response object. In that context `status` sets the status code and requires a call to `send` to actually send the response, while `sendStatus` will set the code and send an empty body for you.

libV2/utils/timestamp.js
tmacro: My reason for abstracting this is that is requires usage of several "magic" numbers. Initially my thoughts were to add constants and do the logic in the places it was needed, but decided that since a wrong conversion could lead to subtle bugs (I've already made this mistake personally several times) that adding a util function made sense. With regards to the use of `<` I can't think of a different way to distinguish timestamps of a particular precision

jonathan-gramain: OK. I still think keeping the timestamp type around (milli or micro, essentially) would be cleaner than guessing based on the value range, but understand your point, not blocking just may be worth rethinking in the future.

libV2/server/API/metrics/listMetrics.js
jonathan-gramain: OK, then let's keep as is.

jonathan-gramain (approved):

libV2/utils/timestamp.js
tmacro: One of the uses for this is the timestamp that is passed into the API which does not come with an attached precision. Currently we only support seconds, but I wanted to lift this as the underlying platform has better resolution now. Warp10 neatly sidesteps this problem by requiring timestamps to be in ISO8601 format so that the precision can be recovered server side while allowing users to use any precision they like. Might be a nice change for API v2

rahulreddy: Moment.js has a helper for this https://momentjs.com/docs/#/parsing/unix-timestamp/

tests/functional/v2/server/testListMetrics.js
```diff
@@ -0,0 +1,150 @@
...
+    .reduce((prev, key) => {
+        prev[key] = 0;
+        return prev;
+    }, {});
+
+async function listMetrics(level, resources, start, end) {
+    const body = {
+        timeRange: [start, end],
+    };
+    body[level] = resources;
```
miniscruff: Inline this?
```suggestion
        [level] = resources,
    };
```

tests/functional/v2/server/testListMetrics.js
```diff
@@ -0,0 +1,150 @@
...
+                    getTs(-360),
+                    getTs(-60),
+                    1000,
+                    testCase.args,
+                );
+                totals = _totals;
+                assert(await ingestEvents(events));
+            });
+
+
```
miniscruff: Extra line?

tests/utils/v2Data/events.js
```diff
@@ -0,0 +1,310 @@
...
+        'listBucket',
+        'putBucketAcl',
+        'putBucketCors',
+        'PutBucketObjectLock',
+        'putBucketReplication',
+        'putBucketVersioning',
+        'putBucketWebsite',
+    ],
+
+
```
miniscruff: double new line

tests/utils/v2Data/events.js
```diff
@@ -0,0 +1,310 @@
...
+    },
+
+    bucket: {
+        user: params => params.user,
+        bucket: params => params.bucket,
+    },
+    obj: objOpFields,
+};
+
+const eventTemplateToOperation = {
```
miniscruff, line 71: this feels like something that should be outside of the test utils, but maybe it is only useful for tests not sure.

tests/utils/v2Data/rand.js
```diff
@@ -0,0 +1,27 @@
+function range(n, step) {
```
miniscruff, line 1: Wait... is this from another project? I swear I remember commenting that this can be a generator before...

tests/utils/v2Data/events.js
```diff
@@ -0,0 +1,310 @@
...
+    in: 0,
+    out: 0,
+};
+
+function updateTotal(total, event) {
+    if (total === undefined) {
+        total = { ..._emptyTotal };
+        total.ops = {};
+    }
+    total.bytes += event.sizeDelta || 0;
```
miniscruff, line 210: Wasn't able to find where event is defined here but what is the default if its not 0 here? is it undefined?

miniscruff (commented):
Mostly small things, a couple questions that I have as well.

tests/utils/v2Data/events.js
tmacro: yes, ops like `putBucketReplication` won't contain a `sizeDelta`

tests/utils/v2Data/events.js
tmacro:
```suggestion
```

tests/utils/v2Data/events.js
tmacro: It's just used for the fake event generation.

tests/functional/v2/server/testListMetrics.js
tmacro:
```suggestion
```

tests/utils/v2Data/rand.js
tmacro: same project, it's just moving places

miniscruff (approved):
Looks good to me but the build is failing as a heads up

PR #580, opened 2020-07-31
ft: S3C-3132 export utapi v2 client

dora-korpar:
A few changes to support v2 pushmetric with cloudserver


libV2/constants.js
```diff
@@ -26,11 +26,14 @@ module.exports = {
         'getBucketAcl',
         'getBucketCors',
         'getBucketLocation',
+        'getBucketObjectLock',
```
tmacro, line 29: I think these are added in one of the open PRs. Depending which merges first we may need to watch for duplicates.

tmacro (approved):

naren-rajendran (approved):

dora-korpar:
@bert-e approve

PR #593, opened 2020-08-04
ft(S3C-3230): Add authentication support

tmacro:


libV2/config/index.js
```diff
@@ -212,6 +212,11 @@ class Config {
         parsedConfig.checkpointSchedule = _loadFromEnv('CHECKPOINT_SCHEDULE', config.checkpointSchedule);
         parsedConfig.snapshotSchedule = _loadFromEnv('SNAPSHOT_SCHEDULE', config.snapshotSchedule);
 
+        const vaultConf = {};
+        vaultConf.host = _loadFromEnv('VAULT_HOST', config.vaultd.host);
+        vaultConf.port = _loadFromEnv('VAULT_PORT', config.vaultd.port);
+        parsedConfig.vaultd = vaultConf;
```
miniscruff: Can inline this
```suggestion
        parsedConfig.vaultd = {
            host: _loadFromEnv('VAULT_HOST', config.vaultd.host),
            port: _loadFromEnv('VAULT_PORT', config.vaultd.port),
        };
```

libV2/server/controller.js
```diff
@@ -107,14 +121,15 @@ class APIController {
...
      */
-    static async _callOperation(handler, request, response) {
-        request.ctx = APIController._buildRequestContext(request);
+    static async _callOperation(handler, request, response, params) {
         try {
-            await handler(request.ctx, APIController._extractParams(request));
+            await handler(request.ctx, params);
         } catch (err) {
             request.logger.error('error during operation', { err });
+            console.log(err);
```
miniscruff: console log left?

libV2/server/middleware.js
```diff
@@ -80,11 +81,41 @@ async function initializeOasTools(spec, app) {
     return promisify(oasTools.initialize)(spec, app);
 }
 
+// eslint-disable-next-line no-unused-vars
+async function authV4Middleware(request, response, params) {
+    // const { params } = request.swagger;
```
miniscruff: commented code

miniscruff (approved):
Looks good from the code side

libV2/errors/errors.json
```diff
@@ -1,5 +1,5 @@
 {
-    "AuthenticationFailed": {
+    "AccessDenied": {
         "code": 403,
         "description": "Authentication failed."
```
jonathan-gramain: `"description": "Access denied"`?

libV2/server/middleware.js
```diff
@@ -80,11 +81,40 @@ async function initializeOasTools(spec, app) {
     return promisify(oasTools.initialize)(spec, app);
 }
 
+// eslint-disable-next-line no-unused-vars
+async function authV4Middleware(request, response, params) {
+    const authHeader = request.headers.authorization;
+    if (!authHeader || !authHeader.startsWith('AWS4')) {
```
jonathan-gramain: could be a minor point, but I think it's more correct to stop comparing after the delimiter (space?) than before, so like `startsWith("AWS4 ")`

libV2/vault.js
```diff
@@ -0,0 +1,144 @@
...
+                try {
+                    const authorizedResources = (res || [])
+                        .reduce((authed, result) => {
+                            if (result.isAllowed) {
+                                // result.arn should be of format:
+                                // arn:scality:utapi:::resourcetype/resource
+                                assert(typeof result.arn === 'string');
+                                assert(result.arn.indexOf('/') > -1);
+                            }
+                            const resource = result.arn.split('/')[1];
```
jonathan-gramain: Does this always work in case `! result.isAllowed`?

tmacro: whoops, those should be inside the conditional

libV2/server/middleware.js
tmacro: I'm not very knowledgeable about what we support wrt authV4, but I was expecting a header in the format `AWS4-HMAC-SHA256`. I don't know if we support other signing algos so it may make sense to compare the entire string? The logic here was mostly lifted from `/lib/Vault.js` and `router/Router.js`

jonathan-gramain: Yes why not comparing against `AWS-HMAC-SHA256[delim]` then, but keeping what is already there in vault code and works is fine too.

tmacro: I've added a `-`

jonathan-gramain (approved):

nicolas2bert (approved):

PR #639, opened 2020-08-19
ft(S3C-3231): Add API version support

tmacro:


libV2/server/legacy.js
```diff
@@ -0,0 +1,114 @@
...
+    module: 'server.legacy',
+});
+
+
+/**
+ * Function to validate a URI component
+ *
+ * @param {string|object} component - path from url.parse of request.url
+ * (pathname plus query) or query from request
+ * @return {string|undefined} If `decodeURIComponent` throws an error,
```
nicolas2bert, line 24: `_checkURIComponent` returns:
- `true` if it is an invalid string or the first invalid key if object type.
- `undefined` if the string or object is valid. 

What about having an `isValidURI` function that returns true if valid and false otherwise.

Here we are returning different types of values which can be very confusing and significantly harder to maintain.

Also, are there tests for it?

tmacro: This class exists because the "legacy" utapi v1 server class isn't exposed. The entire function was lifted from here https://github.com/scality/utapi/blob/018a1fcdb531fda509000fdab73acd4171f851f6/lib/server.js#L48.

PR #642, opened 2020-08-24
ft(S3C-3269): Add TLS support

tmacro:


libV2/config/index.js
```diff
@@ -66,33 +66,46 @@ class Config {
...
         Object.assign(this, parsedConfig);
     }
 
-    static _readFile(path) {
+    static _readFile(path, encoding = 'utf-8') {
+        assert.doesNotThrow(
+            // eslint-disable-next-line no-bitwise
+            () => fs.accessSync(path, fs.F_OK | fs.R_OK),
+            `File not found or unreachable: ${path}`,
+        );
```
miniscruff: What is the benefit of asserting this does not throw? Is it so that it can throw errors as long as its not the file not found error?

tmacro: I was just matching v1 behavior when it loads certificates, but pushed it into a general function. In my experience wrong path/ permissions are the most common errors loading files so I think it makes sense to break out that case.

miniscruff: I do not disagree just wondering the benefits of using `assert.doesNotThrow` or just accessSync. Eg
```suggestion
        fs.accessSync(path, fs.F_OK | fs.R_OK)
```

Note that the error in both cases are nearly identical just that the assertion error wraps the actual message and will alter the callstack ( I believe the node REPL is different ).
image

tmacro: Ah, I see what you mean. Yeah its a little redundant, I'll remove it

miniscruff (approved):

rahulreddy (approved):

PR #647, opened 2020-09-04
ft(S3C-3286): Add reindex task

tmacro:
Most of the line count are yarn.lock changes in the first commit

### DIFF
--- a/libV2/metadata/index.js
+++ b/libV2/metadata/index.js
@@ -1,0 +1,116 @@
+/* eslint-disable no-restricted-syntax */
+const { usersBucket, splitter: mdKeySplitter, mpuBucketPrefix } = require('arsenal').constants;
+const metadata = require('./client');
+const { LoggerContext, logger } = require('../utils');
+const { keyVersionSplitter } = require('../constants');
+
+const moduleLogger = new LoggerContext({
+    module: 'metadata.client',
+});
+
+const PAGE_SIZE = 1000;
+
+function _listingWrapper(bucket, params) {
+    return new Promise(
+        (resolve, reject) => metadata.listObject(
+            bucket,
+            params,
+            logger.newRequestLogger(),
+            (err, res) => {
+                if (err) {
+                    reject(err);
+                    return;
+                }
+                resolve(res);
+            },
+        ),
+    );
+}
+
+function _listObject(bucket, prefix, hydrateFunc) {
+    const listingParams = { prefix, maxKeys: PAGE_SIZE, listingType: 'Basic' };
+    let gt;
+    return {
+        async* [Symbol.asyncIterator]() {
+            while (true) {
+                let res;
+
+                try {
+                    // eslint-disable-next-line no-await-in-loop
+                    res = await _listingWrapper(bucket, { ...listingParams, gt });
+                } catch (error) {
+                    moduleLogger.error('Error during listing', { error });
+                    throw error;
+                }
+
+                for (const item of res) {
+                    yield hydrateFunc ? hydrateFunc(item) : item;
+                }
+
+                if (res.length !== PAGE_SIZE) {
+                    break;
+                }
+
+                gt = res[res.length - 1].key;
+            }
+        },
+    };
+}
+
+function listObjects(bucket) {
+    return _listObject(bucket, '', data => {
+        const { key, value } = data;
+        const [name, version] = key.split(keyVersionSplitter);
+        return {
+            name,
+            version,
+            value: JSON.parse(value),
+        };
+    });
+}
+
+function listBuckets() {
+    return _listObject(usersBucket, '', data => {
+        const { key, value } = data;
+        const [account, name] = key.split(mdKeySplitter);
+        return {
+            account,
+            name,
+            value: JSON.parse(value),
+        };
+    });
+}
+
+async function listMPUs(bucket) {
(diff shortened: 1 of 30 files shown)


tcarmet (approved):

libV2/metadata/index.js
```diff
@@ -0,0 +1,116 @@
+/* eslint-disable no-restricted-syntax */
+const { usersBucket, splitter: mdKeySplitter, mpuBucketPrefix } = require('arsenal/lib/constants');
```
jonathan-gramain: Could be better to access only exposed values from arsenal, so doing instead: `require('arsenal').constants`

libV2/metadata/index.js
```diff
@@ -0,0 +1,116 @@
...
+                    return;
+                }
+                resolve(res);
+            },
+        ),
+    );
+}
+
+function _listObject(bucket, prefix, hydrateFunc) {
+    const listingParams = { prefix, maxKeys: PAGE_SIZE, listingType: 'Basic' };
```
jonathan-gramain, line 31: Due to bug S3C-2987 that is being worked on, it's possible that listing with `Basic` will not be appropriate anymore in the new format. Even if we can stick with it now, it might be good to open a ticket to keep track of this, as I think it will likely require changes once the new format is introduced.

libV2/metadata/index.js
```diff
@@ -0,0 +1,116 @@
...
+    return {
+        async* [Symbol.asyncIterator]() {
+            while (true) {
+                let res;
+
+                try {
+                    // eslint-disable-next-line no-await-in-loop
+                    res = await _listingWrapper(bucket, { ...listingParams, gt });
+                } catch (error) {
+                    moduleLogger.error('Error during listing', { error });
```
jonathan-gramain, line 42: Could be preferable to do `error: error.message` in case `error.message` is a generated field that would then be hidden.

libV2/metadata/index.js
```diff
@@ -0,0 +1,116 @@
...
+                    throw error;
+                }
+
+                for (const item of res) {
+                    yield hydrateFunc ? hydrateFunc(item) : item;
+                }
+
+                if (res.length === 0 || res.length !== PAGE_SIZE) {
```
jonathan-gramain: You could remove the first test here

libV2/tasks/Reindex.js
```diff
@@ -0,0 +1,177 @@
+/* eslint-disable no-restricted-syntax */
+const async = require('async');
+const { mpuBucketPrefix } = require('arsenal/lib/constants');
```
jonathan-gramain: same here about importing from exposed attributes

libV2/tasks/Reindex.js
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+
+    async _execute() {
+        logger.debug('reindexing objects');
+
+        const accountTotals = {};
+        const ignoredAccounts = new Set();
+
+        await async.eachLimit(metadata.listBuckets(), 5, async bucket => {
```
jonathan-gramain, line 120: I'm surprised the `async` module functions can take real async functions returning promises and can be themselves async functions, but I'm not fully aware so it is likely correct!

tests/functional/v2/task/testReindex.js
```diff
@@ -0,0 +1,89 @@
...
+    ops: {},
+    sizeD: 2048,
+    objD: 2,
+    inB: 0,
+    outB: 0,
+};
+
+// eslint-disable-next-line func-names
+describe('Test ReindexTask', function () {
+    this.timeout(1200000);
```
jonathan-gramain: Isn't this timeout too high for the expected run time of this functional test?

libV2/metadata/index.js
tmacro: I picked `Basic` as it made dealing with versioned vs non-versioned object much easier as I can list them using the same API calls. I'll create a ticket to track the update changes if any

libV2/metadata/index.js
tmacro: `error.message` is pulled out if it exists by `moduleLogger` and added as `errmsg`

libV2/tasks/Reindex.js
tmacro: Me too! They were added in 3.x, I believe we mostly use 2.x in other projects

tests/functional/v2/task/testReindex.js
tmacro: yeah this looks like a debug value that made it in

libV2/tasks/Reindex.js
jonathan-gramain: OK, that's what I thought too :)

anurag4DSB (approved):

tests/functional/v2/task/testReindex.js
```diff
@@ -0,0 +1,89 @@
+const assert = require('assert');
+const uuid = require('uuid');
+const { mpuBucketPrefix } = require('arsenal/lib/constants');
```
jonathan-gramain: `require('arsenal').constants`

jonathan-gramain (approved):

tests/functional/v2/task/testReindex.js
tmacro: My editor likes to "fix" them for me :face_with_head_bandage:

PR #651, opened 2020-09-10
ft(S3C-3324): Add counter backend to cache

tmacro:


libV2/cache/backend/memory.js
```diff
@@ -1,10 +1,23 @@
 const schema = require('../schema');
+const constants = require('../../constants');
+
+/**
+ * Returns null iff the value is undefined.
```
miniscruff, line 5:
```suggestion
 * Returns null if the value is undefined.
```

libV2/cache/backend/memory.js
```diff
@@ -1,10 +1,23 @@
...
+ * Returns the passed value otherwise.
+ *
+ * @param {*} value - Any value
+ * @returns {*} - PAssed value or null
```
miniscruff:
```suggestion
 * @returns {*} - Passed value or null
```

tests/unit/v2/cache/testMemoryBackend.js
```diff
@@ -103,4 +103,27 @@ describe('Test cache memory backend', () => {
             assert(res, true);
         });
     });
+
+    it('should update the account size counter', async () => {
+        await Promise.all(testValues.map(event => mem.updateCounters(event)));
```
miniscruff, line 108: I think you can just pass update counters to the map function.
```suggestion
        await Promise.all(testValues.map(mem.updateCounters));
```

tests/unit/v2/cache/testMemoryBackend.js
```diff
@@ -103,4 +103,27 @@ describe('Test cache memory backend', () => {
...
+    });
+
+    it('should update the account size counter base', async () => {
+        await mem.updateAccountCounterBase('imanaccount', 1);
+        const baseKey = schema.getAccountSizeCounterBaseKey('utapi', 'imanaccount');
+        const counterKey = schema.getAccountSizeCounterKey('utapi', 'imanaccount');
+        assert.deepStrictEqual(mem._data, { [baseKey]: 1, [counterKey]: 0 });
+    });
+
+    it('should fetch the account size counter  and base', async () => {
```
miniscruff: double space
```suggestion
    it('should fetch the account size counter and base', async () => {
```

libV2/config/defaults.json
```diff
@@ -5,6 +5,10 @@
         "logLevel": "info",
         "dumpLevel": "error"
     },
+    "redis": {
+        "host": "127.0.0.1",
+        "port": 6379
+    },
```
miniscruff, line 11: Just want to confirm this is intentionally identical to localCache

libV2/cache/backend/redis.js
```diff
@@ -176,6 +177,44 @@ class RedisCache {
...
+            const accountSizeKey = schema.getAccountSizeCounterKey(this._prefix, account);
+            const accountSizeBaseKey = schema.getAccountSizeCounterBaseKey(this._prefix, account);
+            await this._redis.mset(accountSizeKey, 0, accountSizeBaseKey, size);
+            await this._redis.expire(accountSizeBaseKey, constants.counterBaseValueExpiration);
+        } catch (error) {
+            moduleLogger
+                .with({ method: 'updateAccountCounterBase' })
+                .error('error while updating metric counter base', { error });
+            throw error;
+        }
```
miniscruff: I have seen this or something very similar a bit and I think its an alternative method would be to use a decorator. Although javascript decorators are still a proposal you could wrap it yourself easily enough. Consider if say the module logger had a method like this:

```javascript
function errorLogger(fn, description) {
    return function(...params) {
        try {
            fn(...params);
        } catch (error) {
            this.with({method: fn.name })
            this.error(description, { error } );
            throw error
        }
    }
}
```
( would need one for async that returns an async function but you get the point.

Then your function could just be:

```javascript
updateAccountCounterBase = moduleLogger.asyncLogger((account, size) => {
    const accountSizeKey = schema.getAccountSizeCounterKey(this._prefix, account);
    const accountSizeBaseKey = schema.getAccountSizeCounterBaseKey(this._prefix, account);
    await this._redis.mset(accountSizeKey, 0, accountSizeBaseKey, size);
    await this._redis.expire(accountSizeBaseKey, constants.counterBaseValueExpiration);
}, 'error while updating metric counter base')
```
Note that after the decorator proposal you would write
```javascript
@moduleLogger('error while updating metric counter base')
async function(...)
```

If you have used python decorators it seems JS decorators are the same. For reference here is a typescript guide about them https://www.typescriptlang.org/docs/handbook/decorators.html.

libV2/config/defaults.json
tmacro: Yes, in prod `localCache` is a single local redis and `redis` will be a redis sentinel cluster. At some point I'd like to refactor the config format to be a bit more descriptive of the various fields, but for now I'm reusing v1's format and adding where needed.

libV2/cache/backend/redis.js
tmacro: I like the idea of removing the try/catch bolierplate, I'll see what I can do

tmacro: After digging into it, public class fields (this syntax `updateAccountCounterBase = moduleLo....`) is also an open proposal so using decorators would require something like
```javascript
class TestClass {
    constructor() {
        this._data = { val: 'world' };
    }

    _myFunc() {
        console.log(`hello ${this._data.val}`);
    }
}

TestClass.prototype.testmeOut = function () { this._myFunc(); };
```
which is less than ideal, or by moving to a more traditional way of building js classes which would be in contrast to the rest of the project. I instead went with a more traditional wrapper. `getKey/setKey` have been updated to use it, let me know what you think.

libV2/cache/backend/memory.js
tmacro: This ones actually intentional https://www.merriam-webster.com/dictionary/iff

libV2/cache/backend/redis.js
miniscruff: Hmm... dang JS step your game up. Let me see.

libV2/cache/backend/memory.js
miniscruff: oh... TIL

tests/unit/v2/cache/testMemoryBackend.js
tmacro: `updateCounters` would lose its reference to the correct this if not wrapped or prebound with `mem.updateCounters.bind(mem)`

libV2/cache/backend/redis.js
```diff
@@ -43,26 +44,19 @@ class RedisCache {
...
-        } catch (error) {
-            moduleLogger
-                .with({ method: 'getKey' })
-                .error('error fetching key from redis', { key });
-            throw error;
-        }
+        return moduleLogger
+            .with({ method: 'getKey' })
+            .logAsyncError(() => this._redis.get(key),
+                'error fetching key from redis', { key });
```
miniscruff, line 50: I am not sure where `{ key }` is used, the log async error method only has two parameters.

tmacro: ah, it's supposed to be passed along like the other methods

libV2/cache/backend/redis.js
```diff
@@ -43,26 +44,19 @@ class RedisCache {
...
-            const res = await this._redis.set(key, value);
-            return res === 'OK';
-        } catch (error) {
-            moduleLogger
-                .with({ method: 'setKey' })
-                .error('error setting key in redis', { key });
-            throw error;
-        }
+        return moduleLogger
+            .with({ method: 'setKey' })
```
miniscruff, line 55: because you need to create a separate anonymous function we also can't rely on the name of this function from `func.name`... all kinda lame. We could also have automated logging the params from `...params` in the decorator but just seems like a burden without it.

libV2/cache/backend/redis.js
miniscruff: Well your implementation seems like a good alternative but doesn't seem dramatically better than try/catch so I am not sure it is worth going against all the other parts that use try/catch. Maybe if decorators are approved we can make some sweet wrappers ( I love decorators from python ). Looks good to me either way though.

miniscruff (approved):

miniscruff (approved):

anurag4DSB (approved):

anurag4DSB (approved):

PR #671, opened 2020-09-22
bf(S3C-1997): Fix ioredis failover for sentinels

tmacro:


libV2/redis.js
```diff
@@ -0,0 +1,160 @@
...
+            this._redis.off('connect', this._onConnect);
+            this._redis.off('ready', this._onReady);
+            this._redis.off('error', this._onError);
+        }
+        this._isConnected = false;
+        this._isReady = false;
+        this._redis = new IORedis(this._redisOptions);
+        this._redis.on('connect', this._onConnect.bind(this));
+        this._redis.on('ready', this._onReady.bind(this));
+        this._redis.on('error', this._onError.bind(this));
```
anurag4DSB, line 74: Just curious to learn. What additional functionality does .bind bring?

tmacro: Without `.bind`, when the handlers are called they would lose their reference to the correct `this`. Normally I would just wrap it in an anonymous function  like `() => this._onReady()` to achieve the same thing, however doing so would would mean that the anonymous function is registered rather than the instance method which would prevent me from using `.off` to remove them.

libV2/redis.js
```diff
@@ -0,0 +1,160 @@
...
+        this._inFlightTimeouts = this._useTimeouts ? {} : null;
+        this._runningRedisProbe = null;
+        this._isConnected = false;
+        this._isReady = false;
+    }
+
+    connect(callback) {
+        this._initClient(false);
+        if (callback) {
+            callback();
```
jonathan-gramain: I'd recommend using `process.nextTick()` to avoid calling synchronously the callback, with potential side effects.

lib/UtapiClient.js
```diff
@@ -7,9 +7,25 @@ const werelogs = require('werelogs');
 const { errors } = require('arsenal');
 const Datastore = require('./Datastore');
 const { generateKey, generateCounter, generateStateKey } = require('./schema');
-const redisClient = require('../utils/redisClient');
+const RedisClientv2 = require('../libV2/redis');
 const member = require('../utils/member');
 
+function redisClientv2(conf, log) {
```
naren-rajendran: May be add jsdoc for `redisClientv2`, it will be easier to check what's required in `conf`!

lib/UtapiReplay.js
```diff
@@ -5,8 +5,20 @@ const { scheduleJob } = require('node-schedule');
 const werelogs = require('werelogs');
 const UtapiClient = require('./UtapiClient');
 const Datastore = require('./Datastore');
-const redisClient = require('../utils/redisClient');
 const safeJsonParse = require('../utils/safeJsonParse');
+const RedisClientv2 = require('../libV2/redis');
+
+function redisClientv2(conf) {
```
naren-rajendran: Same here, jsdoc!

libV2/redis.js
```diff
@@ -0,0 +1,160 @@
...
+const moduleLogger = new LoggerContext({
+    module: 'redis',
+});
+
+const COMMAND_TIMEOUT = 10000;
+const CONNECTION_TIMEOUT = 30000;
+
+
+class RedisClient extends EventEmitter {
+    constructor(options) {
```
naren-rajendran, line 28: Same here, jsDoc comments!

libV2/utils/func.js
```diff
@@ -0,0 +1,20 @@
+const { callbackify } = require('util');
```
naren-scality, line 1: Should this util be in arsenal?

libV2/utils/func.js
```diff
@@ -0,0 +1,20 @@
...
+
+/**
+ *  Convenience function to handle "if no callback then return a promise" pattern
+ *
+ * @param {Function} asyncFunc - asyncFunction to call
+ * @param {Function|undefined} callback - optional callback
+ * @returns {Promise|undefined} - returns a Promise if no callback is passed
+ */
+function asyncOrCallback(asyncFunc, callback) {
```
naren-rajendran, line 10: Should this util be in arsenal?

libV2/utils/index.js
```diff
@@ -1,9 +1,11 @@
 const log = require('./log');
 const shard = require('./shard');
 const timestamp = require('./timestamp');
+const func = require('./func');
 
 module.exports = {
     ...log,
     ...shard,
     ...timestamp,
+    ...func,
```
naren-rajendran, line 10: Should these be destructured here?

naren-rajendran: I think it would be easier to read if it is `util.func.asyncOrCallback` than `util.asyncOrCallback` no?

libV2/utils/func.js
tmacro: Is there a need in other projects?

libV2/utils/index.js
tmacro: `utils` is normally used like `const { LoggerContext, asyncOrCallback } = require('./utils');` so I think it make sense to export them under a single namespace. Additionally it would make `func` different than the other submodules and I like to keep things consistent.

libV2/utils/func.js
naren-rajendran: could be, but this can be a chore/improvement for later.

utils/redisClientv2.js
```diff
@@ -0,0 +1,30 @@
+const RedisClient = require('../libV2/redis');
```
naren-scality, line 1: I was wondering where this was :)

utils/redisClientv2.js
```diff
@@ -0,0 +1,30 @@
...
+
+    client.connect();
+    client.on('error', err => log.trace('error with redis client', {
+        error: err,
+    }));
+    return client;
+}
+
+
+module.exports = redisClientv2;
```
naren-rajendran, line 30: I was wondering where this file was :)

naren-rajendran (approved):
lgtm

libV2/redis.js
```diff
@@ -0,0 +1,170 @@
...
+        this._isConnected = true;
+        this.emit('connect');
+    }
+
+    _onReady() {
+        this._isReady = true;
+        this.emit('ready');
+    }
+
+    // eslint-disable-next-line class-methods-use-this
```
jonathan-gramain: Is this warning removal needed?

libV2/redis.js
```diff
@@ -0,0 +1,170 @@
...
+    _onError(error) {
+        moduleLogger.error('error connecting to redis', { error });
+        this.emit('error', error);
+    }
+
+
+    // eslint-disable-next-line class-methods-use-this
```
jonathan-gramain: this one too

libV2/redis.js
```diff
@@ -0,0 +1,170 @@
...
+    _createCommandTimeout() {
+        const timerId = uuid.v4();
```
jonathan-gramain: I don't think you need a `timerId`, you could instead use a `Set()` containing your timers, and call `has(timer)` to check the presence and eventually `delete(timer)` providing the timer object itself. Not very important, up to you if you want to do the change or not.

libV2/redis.js
```diff
@@ -0,0 +1,170 @@
...
+        if (!this._useTimeouts) {
+            // If timeouts are disabled simply return the Promise
+            return funcPromise;
+        }
+
+        const { timeout, cancel } = this._createCommandTimeout();
+
+        try {
+            const resp = await Promise.race([funcPromise, timeout]);
+            cancel();
```
jonathan-gramain: All calls to `cancel` including `this.once('timeout', cancel)` I believe might be replaced by a single unconditional `cancel()` call in a `finally {}` block, making sure we don't forget to cancel the timer in some code path (with simpler code).

Also for readability, `cancel` might be a bit confusing as it does not tell what it is canceling: `stopTimeout` or `cancelTimeout` could be better.

tests/functional/cron/testReplay.js
```diff
@@ -223,7 +229,7 @@ function checkAllMetrics(cb) {
     }), err => cb(err));
 }
 
-describe('Replay', () => {
+describe.skip('Replay', () => {
```
jonathan-gramain: Why is this test skipped?

libV2/redis.js
tmacro: They should make it a failure if a flag to turn something off doesn't change anything. I do this all the time

libV2/redis.js
tmacro: I always forget that sets in js don't suffer from the same drawbacks as python's (set items need to be hashable so you can't add a random object to a set)

tests/functional/cron/testReplay.js
tmacro: it shouldn't be

libV2/redis.js
tmacro: I've reworked  this including to include the set changes as well, but I'm on the fence about `this.once('timeout', cancel)`. The intention is to cause concurrent requests to also timeout if any one of them does.

jonathan-gramain: Ah, isn't it better to have each request timeout independently from the other? Just wondering, otherwise if you want to keep the current behavior no problem.

jonathan-gramain (approved):

PR #678, opened 2020-09-23
bf(S3C-1997): Fix ioredis failover for sentinels

tmacro:
This is mostly a direct port of the redis sentinel fix, but because `7.6` was released before utapi v2 some small module need to be ported as well. Additionally a bump to eslint and its peers is needed so that async/await code can be parsed. As this bump also brings large amounts of rule changes several rules have been disabled at the project level.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,1 +1,31 @@
-{ "extends": "scality" }
+{
+    "extends": "scality",
+    "rules": {
+        "no-underscore-dangle": "off",
+        "object-property-newline": "off",
+        "indent": "off",
+        "object-curly-newline": "off",
+        "operator-linebreak": "off",
+        "no-mixed-operators": "off",
+        "class-methods-use-this": "off",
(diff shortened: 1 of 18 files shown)


rahulreddy (approved):

.eslintrc
```diff
@@ -1 +1,32 @@
...
+        "no-plusplus": "off",
+        "global-require": "off",
+        "symbol-description": "off",
+        "prefer-spread": "off",
+        "import/no-extraneous-dependencies": "off",
+        "lines-around-directive": "off",
+        "no-undef-init": "off",
+        "no-bitwise": "off",
+        "operator-assignment": "off",
+    	"implicit-arrow-linebreak" : "off"
```
miniscruff:
```suggestion
    	  "implicit-arrow-linebreak" : "off"
```
To match

miniscruff (approved):

PR #695, opened 2020-09-29
bf(S3C-3307): Handle deleted buckets in reindex

tmacro:


lib/reindex/s3_bucketd.py
```diff
@@ -48,6 +48,14 @@ def inner(*args, **kwargs):
         return urllib.parse.quote(val.encode('utf-8'))
     return inner
 
+def as_set(func):
```
miniscruff: I recommend using functools.wraps for all decorators as it handles all the extra bits plain decorators do not.
https://docs.python.org/3.8/library/functools.html#functools.wraps

lib/reindex/s3_bucketd.py
```diff
@@ -314,6 +322,11 @@ def update_redis(client, resource, name, obj_count, total_size):
     client.set(obj_count_key + ':counter', obj_count)
     client.set(total_size_key + ':counter', total_size)
 
+@as_set
```
miniscruff: Not really sure making a generator method return a set this way is really ideal.
You can already turn generators like this into sets by just using `set(gen_func(...))` and it removes the option of using this as a normal generator if needed.

lib/reindex/s3_bucketd.py
tmacro: I've removed the decorator per your other comment, but what are the extra bits `wraps` preserves? Docs seem to indicate the docstring and function name are saved, is it only metadata or have I been avoiding a massive footgun by chance all these years?

miniscruff: I think its just metadata like docstring and name, I think it does impact special cases like open API spec tools or sqlalchemy will sometimes use that data. It might cause an issue if you stack decorators and they aren't wrapped, then the metadata gets real fuzzy. But it is a relatively small footgun at most, I mention it mostly as a safety thing. And not everyone knows about functools ( itertools is also great ).

lib/reindex/s3_bucketd.py
```diff
@@ -350,3 +356,26 @@ def log_report(resource, name, obj_count, total_size):
...
             log_report('accounts', userid, report['obj_count'], report['total_size'])
         pipeline.execute()
+
+    observed_accounts = set(account_reports.keys())
+    recorded_accounts = set(get_resources_from_redis(redis_client, 'accounts'))
+    recorded_buckets = set(get_resources_from_redis(redis_client, 'buckets'))
+
+    # Stale accounts and buckets are ones that do not appear in the listing, but have recorded values
+    stale_accounts = recorded_accounts.difference(observed_accounts)
+    print('Found %s stale accounts' % len(stale_accounts))
```
miniscruff: Small preference thing but not a huge fan of %s syntax and I believe python even recommends not using it now. But depending on your version it might still be recommended.
Anyway, f-strings are my preference if you are py3.6+ if not, you can just use the fact that print accepts any arguments and will append a space between then.
```suggestion
    print('Found', len(stale_accounts), 'state accounts')
```

miniscruff: oooh... unfortunate... I was writing the comment as your update came in... LOL
Well if log.info supports the same thing it might still apply but meh.

tmacro: I'm not a big fan of the `%` syntax either but we're stuck with 3.4 for this sadly

miniscruff: oh ew 3.4, that one is end of life already. I am not even sure if string.format was in that or not, that might of been 3.5.

lib/reindex/s3_bucketd.py
```diff
@@ -350,3 +356,26 @@ def log_report(resource, name, obj_count, total_size):
...
+
+    # Stale accounts and buckets are ones that do not appear in the listing, but have recorded values
+    stale_accounts = recorded_accounts.difference(observed_accounts)
+    _log.info('Found %s stale accounts' % len(stale_accounts))
+    for chunk in chunks(stale_accounts, ACCOUNT_UPDATE_CHUNKSIZE):
+        pipeline = redis_client.pipeline(transaction=False) # No transaction to reduce redis load
+        for account in chunk:
+            update_redis(pipeline, 'accounts', account, 0, 0)
+            log_report('accounts', account, 0, 0)
+        pipeline.execute()
```
miniscruff, line 372: Unrelated but I think it would be sweet if pipeline was a context manager so you could use `with` like so:
```suggestion
        with redis_client.pipeline(transaction=False) as pipeline:
            for account in chunk:
                update_redis(pipeline, 'accounts', account, 0, 0)
                log_report('accounts', account, 0, 0)
```

miniscruff (approved):

lib/reindex/s3_bucketd.py
```diff
@@ -350,3 +356,26 @@ def log_report(resource, name, obj_count, total_size):
...
+
+    stale_buckets = recorded_buckets.difference(observed_buckets)
```
ilkescality, line 374: I was unaware this exists in python (`difference`). neat.

ilkescality (approved):

lib/reindex/s3_bucketd.py
miniscruff: I think you can also use the `-` operator but tbh the named method works better.

PR #730, opened 2020-10-07
ft(S3C-3418): Change wire format to json for warp10 ingestion

tmacro:
Drop our bespoke packing into warpscript syntax and switch to JSON. I'm choosing this over extending the existing solution as we don't need to maintain any custom code.


jonathan-gramain (approved):

rahulreddy (approved):

ilkescality (approved):
LGTM (noticed some linter failures, but probably minor)

PR #779, opened 2020-10-21
bf(S3C-3447): Switch to node-schedule

tmacro:
Changed to bring dependecy inline with other projects. Additionally solves S3C-3339 in the process


rahulreddy (approved):

libV2/tasks/BaseTask.js
```diff
@@ -1,4 +1,4 @@
-const cron = require('node-cron');
+const cron = require('node-schedule');
```
miniscruff, line 1: should we rename the variable?

rachedbenmustapha: It's still a cron implementation :)

miniscruff: I mean... true. :laughing:

miniscruff (approved):

PR #812, opened 2020-10-28
ft(S3C-3484): Add Sensision and extend warp10 tunables

tmacro:
- Adds static.tokens from the federation image
- Adds a disabled by default sensision metrics collector
- Exposes some new tunables for warp10
- Adds some missing flags used in the upstream warp10 init

### DIFF
--- a/images/warp10/s6/cont-init.d/20-install-config
+++ b/images/warp10/s6/cont-init.d/20-install-config
@@ -10,5 +10,5 @@ for path in $WARP10_CONF_TEMPLATES/*; do
 done
 
 echo "Installing sensision config"
-cp ${SENSISION_HOME}/templates/sensision.template $SENSISION_DATA_DIR/conf/sensision.conf
+cp ${SENSISION_HOME}/templates/sensision.template ${SENSISION_DATA_DIR}/conf/sensision.conf
 cp ${SENSISION_HOME}/templates/log4j.properties.template ${SENSISION_DATA_DIR}/conf/log4j.properties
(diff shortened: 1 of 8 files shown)


jonathan-gramain (approved):

images/warp10/s6/cont-init.d/20-install-config
```diff
@@ -1,9 +1,14 @@
...
     if [ ! -f "$WARP10_DATA_DIR/conf/$name" ]; then
         cp "$path" "$WARP10_DATA_DIR/conf/$name"
         echo "Copied $name to $WARP10_DATA_DIR/conf/$name"
     fi
-done
\ No newline at end of file
+done
+
+echo "Installing sensision config"
+cp ${SENSISION_HOME}/templates/sensision.template $SENSISION_DATA_DIR/conf/sensision.conf
```
miniscruff: just this one var doesn't have braces, not sure if it's intentional or not.

```suggestion
cp ${SENSISION_HOME}/templates/sensision.template ${SENSISION_DATA_DIR}/conf/sensision.conf
```

images/warp10/static.tokens
```diff
@@ -0,0 +1,9 @@
+token.write.0.name=writeTokenStatic
+token.write.0.producer=42424242-4242-4242-4242-424242424242
```
miniscruff, line 2: is this a test file or example file? Or does all 42's actually work?

tmacro: I just reused upstream's from here https://github.com/senx/warp10-docker/blob/master/predictible-tokens-for-ci/ci.tokens
AFAIK that's a valid, if highly unlikely, uuid.

miniscruff: oh are these just for CI purposes?

miniscruff (approved):

images/warp10/static.tokens
tmacro: not only, we use a static token for deployment as well. Different Dockerfile though.

PR #821, opened 2020-11-09
bf(S3C-3514): Add missing call to responseLoggerMiddleware in errorMiddleware

tmacro:
`responseLoggerMiddleware` is moved upward so it defines before `errorMiddleware` to avoid a linting error

### DIFF
--- a/libV2/server/middleware.js
+++ b/libV2/server/middleware.js
@@ -44,6 +44,17 @@ function loggerMiddleware(req, res, next) {
     return next();
 }
 
+function responseLoggerMiddleware(req, res, next) {
+    const info = {
+        httpCode: res.statusCode,
+        httpMessage: res.statusMessage,
+    };
+    req.logger.end('finished handling request', info);
+    if (next !== undefined) {
+        next();
+    }
+}
+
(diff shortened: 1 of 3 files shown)


naren-rajendran (approved):

alexanderchan-scality (approved):

libV2/server/middleware.js
```diff
@@ -70,15 +81,7 @@ function errorMiddleware(err, req, res, next) {
...
-}
-
-function responseLoggerMiddleware(req, res, next) {
-    const info = {
-        httpCode: res.statusCode,
-        httpMessage: res.statusMessage,
-    };
-    req.logger.end('finished handling request', info);
-    return next();
+    responseLoggerMiddleware(req, res);
```
miniscruff, line 84: should this pass in next?

tmacro: We stop calling `next` in `errorMiddleware` because we write out all our responses there and want to halt the chain. I chose to make the `next` param optional for `responseLoggerMiddleware` rather than passing a no-op in.

miniscruff (approved):

tmacro:
@bert-e ping

PR #828, opened 2020-11-11
impr(S3C-3520): Remove intermediate hex encoding for warp10 datapoints

tmacro:


warpscript/utapi/decodeEvent.mc2
```diff
@@ -17,18 +17,22 @@ message Event {
 '>
 PROTOC 'proto' STORE
 <%
-  HEX-> !$proto 'Event' PB->
+  'iso8859-1' ->BYTES
```
jonathan-gramain, line 20: why `iso8859-1`? I'm thinking there should be an appropriate "binary" encoding (meaning no encoding at all) for binary data. If you're sure of what this does, feel free to ignore, but I find it suspicious. Or if it's pure ASCII, there must be an "ascii" encoding available to use.

tmacro: warp10 internally encodes binary values as `iso8859-1` as stores them as a "string", but currently doesn't have support in `FETCH` (what we use to load data) to auto convert it back so we need to do it manually

vrancurel (approved):

jonathan-gramain (approved):

PR #853, opened 2020-12-11
bf(S3C-3505): Add support for --bucket flag to s3_reindex.py

tmacro:
re add functionality lost in the rewrite

### DIFF
--- a/lib/reindex/s3_bucketd.py
+++ b/lib/reindex/s3_bucketd.py
@@ -32,7 +32,7 @@ def get_options():
     parser.add_argument("-n", "--sentinel-cluster-name", default='scality-s3', help="Redis cluster name")
     parser.add_argument("-s", "--bucketd-addr", default='http://127.0.0.1:9000', help="URL of the bucketd server")
     parser.add_argument("-w", "--worker", default=10, help="Number of workers")
-    parser.add_argument("-b", "--bucket", default=False, help="Bucket to be processed")
+    parser.add_argument("-b", "--bucket", default=None, help="Bucket to be processed")
     return parser.parse_args()
 
 def chunks(iterable, size):
@@ -119,7 +119,7 @@ class BucketDClient:
             else:
                 is_truncated = len(payload) > 0
 
-    def list_buckets(self):
+    def list_buckets(self, name = None):
 
         def get_next_marker(p):
             if p is None:
@@ -135,8 +135,14 @@ class BucketDClient:
             buckets = []
             for result in payload['Contents']:
                 match = re.match("(\w+)..\|..(\w+.*)", result['key'])
-                buckets.append(Bucket(*match.groups()))
-            yield buckets
+                bucket = Bucket(*match.groups())
+                if name is None or bucket.name == name:
+                    buckets.append(bucket)
+            if buckets:
+                yield buckets
+                if name is not None:
+                    # Break on the first matching bucket if a name is given
+                    break
 
 
     def list_mpus(self, bucket):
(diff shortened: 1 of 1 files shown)


lib/reindex/s3_bucketd.py
```diff
@@ -135,8 +135,14 @@ def get_next_marker(p):
             buckets = []
             for result in payload['Contents']:
                 match = re.match("(\w+)..\|..(\w+.*)", result['key'])
-                buckets.append(Bucket(*match.groups()))
-            yield buckets
+                bucket = Bucket(*match.groups())
+                if name is None or bucket.name == name:
+                    buckets.append(bucket)
+            if len(buckets):
```
miniscruff: You could just do `if buckets:`

lib/reindex/s3_bucketd.py
```diff
@@ -349,33 +355,42 @@ def log_report(resource, name, obj_count, total_size):
...
-        pipeline = redis_client.pipeline(transaction=False) # No transaction to reduce redis load
-        for userid, report in chunk:
-            update_redis(pipeline, 'accounts', userid, report['obj_count'], report['total_size'])
-            log_report('accounts', userid, report['obj_count'], report['total_size'])
-        pipeline.execute()
-
-    observed_accounts = set(account_reports.keys())
-    recorded_accounts = set(get_resources_from_redis(redis_client, 'accounts'))
     recorded_buckets = set(get_resources_from_redis(redis_client, 'buckets'))
+    if options.bucket is None:
```
miniscruff, line 362: an alternative could be to use `if not options.bucket` but it depends on how you might handle say empty strings or 0 in the case of numbers. 

Edit: Just to clarify `if not ""` is true while `if "" is None` is false as my reasoning to one or the other.

lib/reindex/s3_bucketd.py
```diff
@@ -349,33 +355,42 @@ def log_report(resource, name, obj_count, total_size):
...
+        stale_buckets = recorded_buckets.difference(observed_buckets)
+    elif len(observed_buckets) == 0 and options.bucket in recorded_buckets:
```
miniscruff: I feel like I see `not observed_buckets` more than comparing len to 0, will have the same results though.

lib/reindex/s3_bucketd.py
```diff
@@ -349,33 +355,42 @@ def log_report(resource, name, obj_count, total_size):
...
-        pipeline = redis_client.pipeline(transaction=False) # No transaction to reduce redis load
-        for account in chunk:
-            update_redis(pipeline, 'accounts', account, 0, 0)
-            log_report('accounts', account, 0, 0)
-        pipeline.execute()
-
-    stale_buckets = recorded_buckets.difference(observed_buckets)
     _log.info('Found %s stale buckets' % len(stale_buckets))
     for chunk in chunks(stale_buckets, ACCOUNT_UPDATE_CHUNKSIZE):
         pipeline = redis_client.pipeline(transaction=False) # No transaction to reduce redis load
```
miniscruff, line 372: would be it possible / more efficient to declare the pipeline before the for loop instead of in each iteration?

lib/reindex/s3_bucketd.py
tmacro: js habits leaking over...

lib/reindex/s3_bucketd.py
tmacro: I added a bit of L#337 to catch empty strings, but I'm gonna leave this as `is None` as I prefer the strict checking of `is` over coercing a boolean from the arg.

miniscruff: ok with the extra check that makes sense

miniscruff (approved):

lib/reindex/s3_bucketd.py
tmacro: I couldn't find anything in the redis-py docs to suggest one can reuse pipelines, the intent here is to batch commands to redis. (pipeline caches commands client-side until `.execute()` is called)

ilkescality (approved):

PR #864, opened 2020-12-31
ft(S3C-3771): Add support for Utapiv2 client to communicate over TLS

tmacro:


naren-rajendran (approved):

anurag4DSB (approved):

lib/UtapiClient.js
```diff
@@ -126,7 +126,7 @@ class UtapiClient {
         this.enabledOperationCounters = [];
         this.disableClient = true;
 
-        if (config) {
+        if (config && !config.disableClient) {
```
anurag4DSB, line 129: should it be this.disbleClient?

tmacro: `config.disableClient` is correct. The intent is to allow config to be passed while not enabling it as cloudserver always instantiates a client at startup and relies on the config being undefined to disable it. Context is that I need to pass the tls certs into the v2 client and would like to keep the logic on cloudservers side simple

anurag4DSB: Cool

PR #867, opened 2021-01-07
ft(S3C-3524): upgrade warp10 to 2.7.2

tmacro:


miniscruff (approved):
Random aside, does dependabot support upgrading base images or some other tool?

tmacro:
not that I'm aware of

ilkescality (approved):

rahulreddy (approved):

rahulreddy:
@miniscruff I think it could if there is a way to declare our versions as python's requirements.txt or yum or some other standardized way package managers use to manage dependencies.

rahulreddy:
> @miniscruff I think it could if there is a way to declare our versions as python's requirements.txt or yum or some other standardized way package managers use to manage dependencies.

Scratch that, I thought that I was in Federation for a second.
@tmacro You are bumping the version in Federation as well right? This image is used by Utapi project specifically.

tmacro:
@rahulreddy yes

PR #879, opened 2021-01-15
ft(S3C-3707): Add disk usage soft limit task

tmacro:

PR #882, opened 2021-01-21
Enable and fix functional tests

tmacro:
I found a stray `.only` when writing tests for #879 causing all but a single functional test suite to be disabled. The changes grew a bit large so I've split it into a separate PR here.

### DIFF
--- a/tests/functional/v2/task/testCreateCheckpoint.js
+++ b/tests/functional/v2/task/testCreateCheckpoint.js
@@ -5,7 +5,7 @@ const { Warp10Client } = require('../../../../libV2/warp10');
 const { convertTimestamp } = require('../../../../libV2/utils');
 const { CreateCheckpoint } = require('../../../../libV2/tasks');
 
-const { generateCustomEvents, protobuf } = require('../../../utils/v2Data');
+const { generateCustomEvents, fetchRecords } = require('../../../utils/v2Data');
 
 const _now = Math.floor(new Date().getTime() / 1000);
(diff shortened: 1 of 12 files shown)


tests/functional/v2/task/testCreateCheckpoint.js
```diff
@@ -16,6 +16,33 @@ const checkpointLabelToLevel = {
     usr: 'users',
 };
 
+// async function fetchRecords(warp10, className, labels, params) {
```
miniscruff: large commented block

tests/functional/v2/task/testRepair.js
```diff
@@ -113,16 +119,18 @@ describe('Test Repair', function () {
...
+            { node: prefix },
+            { end: getTs(1), count: -1 },
+            '@utapi/decodeRecord',
+        );
         assert.strictEqual(series.length, 3);
         assertResults(totals, series);
 
         await repairTask._execute(getTs(0));
-        results = await warp10.fetch({
+        const results = await warp10.fetch({
```
jonathan-gramain: not possible or no benefit in using `fetchRecords` for this one?

jonathan-gramain (approved):

tests/functional/v2/task/testRepair.js
tmacro: I didn't convert it as we're only concerned about the length and don't need to decode, but I'll change it for consistency

anurag4DSB (approved):

PR #888, opened 2021-01-27
ft(S3C-3609): Add hard disk limit

tmacro:


libV2/tasks/DiskUsage.js
```diff
@@ -166,6 +206,18 @@ class MonitorDiskUsage extends BaseTask {
...
         }
+
+        if (this._hardLimit !== undefined) {
+            const shouldLock = this._checkHardLimit(size, this.nodeId);
+            if (shouldLock) {
+                moduleLogger.error('hard limit exceeded, disabling writes to warp 10', { nodeId: this.nodeId });
+                await this._disableWarp10Updates();
+            } else {
+                moduleLogger.error('usage below hard limit, ensuring writes to warp 10 are enabled',
+                    { nodeId: this.nodeId });
```
vrancurel, line 245: At some point we said that we should not re-enable writes automatically (since we lost events, and repair is needed, etc.)

images/warp10/s6/cont-init.d/10-generate-hash-seeds
```diff
@@ -8,4 +8,5 @@ WARP10_SECRETS="$WARP10_CONFIG_DIR/00-secrets.conf"
 if [ ! -f "$WARP10_SECRETS" ]; then
     cp "$WARP10_CONF_TEMPLATES/00-secrets.conf.template" "$WARP10_SECRETS"
     python "${WARP10_HOME}/etc/generate_crypto_key.py" "$WARP10_SECRETS"
+    echo "warp10.manager.secret = scality" >> $WARP10_SECRETS
```
vrancurel, line 11: Why not calling `generate_crypto_key.py` here ?

tmacro: Internally `generate_crypto_key.py` is using a regex to insert random values for the secrets.
There's 2 problems with this:
- we need the management secret to be known to the hard limit process and random keys makes this difficult at deploy
- The line for `warp10.manager.secret` doesn't contain the template value `generate_crypto_key.py` uses to insert

libV2/tasks/DiskUsage.js
tmacro: Ah I forgot about that. #899 removes the technical barriers to automatic unlocking as far as lost events/consistency, though I think there may have been an ops reason as well? Either way I'll add an manual unlock flag to the cli

vrancurel: if you say so

vrancurel: No risk of inconsistency if let say, 2 nodes were locked for 2 days ?

images/warp10/s6/cont-init.d/10-generate-hash-seeds
vrancurel: ok, we might solve this later, plz create a ticket.

vrancurel (approved):

libV2/tasks/DiskUsage.js
tmacro: Under the new system any events that were destined to be written to the locked warp 10s would have been written to one of the available nodes and processed there. The snapshot/checkpoint/repair processes already allow for arbitrary lengths of time in between them so from their perspective a locked warp 10 would look the same as an idle cluster.

tmacro: Not counting the lag in replication between warp 10s it would introduce of course

libV2/tasks/DiskUsage.js
```diff
@@ -100,7 +118,58 @@ class MonitorDiskUsage extends BaseTask {
...
+        const hardLogger = moduleLogger.with({
+            size,
+            sizeHuman: formatDiskSize(size),
+            hardPercentage,
+            hardLimit: this._hardLimit,
+            hardLimitHuman,
+            nodeId,
+        });
+
+        const msg = `Using ${hardPercentage * 100}% of the ${hardLimitHuman} soft limit on ${nodeId}`;
```
jonathan-gramain: `of the ${hardLimitHuman} soft limit` -> `of the ${hardLimitHuman} hard limit`

ilkescality (approved):

tests/functional/hardLimit/testHardLimit.js
```diff
@@ -0,0 +1,50 @@
...
+
+    beforeEach(async () => {
+        path = `/tmp/diskusage-${uuid.v4()}`;
+        task = new MonitorDiskUsage();
+        await task.setup();
+        task._path = path;
+        task._enabled = true;
+    });
+
+    it('should trigger a database unlock if below the limit', async () => {
```
jonathan-gramain: very minor: could be more logical to put the lock test before the unlock test since unlock happens after lock

libV2/tasks/DiskUsage.js
```diff
@@ -165,6 +234,18 @@ class MonitorDiskUsage extends BaseTask {
...
+
+        if (this._hardLimit !== undefined) {
+            const shouldLock = this._checkHardLimit(size, this.nodeId);
+            if (shouldLock) {
+                moduleLogger.error('hard limit exceeded, disabling writes to warp 10', { nodeId: this.nodeId });
+                await this._disableWarp10Updates();
+            } else {
+                moduleLogger.error('usage below hard limit, ensuring writes to warp 10 are enabled',
+                    { nodeId: this.nodeId });
+                await this._enableWarp10Updates();
```
jonathan-gramain, line 246: Wondering if we should not keep a state if we are currently above the hard limit or not, in order to avoid the overhead of re-enabling updates every time we happen to be below the hard limit (i.e. executing the `DROP DROP 'scality' UPDATEON` code), even if we did not reach the hard limit yet? Not sure I understand the logic, just asking.

tmacro: I thought about this, but decided that a simpler mechanism for the first version was ok as the overhead is minimal

jonathan-gramain (approved):

PR #889, opened 2021-01-27
Improvement/s3 c 3800 rework failover

tmacro:


libV2/tasks/BaseTask.js
```diff
@@ -104,6 +94,22 @@ class BaseTask extends Process {
     async _join() {
         return this._cache.disconnect();
     }
+
+    withWarp10(func, onError) {
+        return iterIfError(this._warp10Clients, func, error => {
+            const {
+                name, code, message, stack,
+            } = error;
+            logger.error('error during warp 10 request', {
```
jonathan-gramain: Thinking out loud if all errors should be logged in error, or if some errors are "normal" and should be logged in debug/info

libV2/utils/func.js
```diff
@@ -30,7 +30,39 @@ function comprehend(data, func) {
...
+ *
+ * onError, if passed, is called on every error thrown by func;
+ *
+ * @param {Array} items - items to iterate
+ * @param {AsyncFunction} func - function to apply to each item
+ * @param {Function|undefined} onError - optional function called if an error is thrown
+ * @returns {*} -
+ */
+async function iterIfError(items, func, onError) {
+    // console.log(items)
```
jonathan-gramain: Debugging comments should be removed

jonathan-gramain (approved):

libV2/tasks/BaseTask.js
tmacro: All cases I can think of meet the "error" qualification, though I can see value in being able to silence this. What do you think of changing it to work as a default value for `onError`? That should allow opting out on a case by case basis.

libV2/tasks/BaseTask.js
```diff
@@ -104,6 +94,22 @@ class BaseTask extends Process {
...
+                error: {
+                    name, code, errmsg: message, stack: name !== 'RequestError' ? stack : undefined,
+                },
+            });
+            if (onError) {
+                onError(error);
+            }
```
tmacro:
```suggestion
            if (onError) {
                onError(error);
            } else {
                const {
                    name, code, message, stack,
                } = error;
                logger.error('error during warp 10 request', {
                    error: {
                        name, code, errmsg: message, stack: name !== 'RequestError' ? stack : undefined,
                    },
                });
            }
```

naren-rajendran (approved):

PR #896, opened 2021-02-02
bugfix: S3C-3830 bump warp10

rahulreddy:


naren-rajendran (approved):

miniscruff (approved):

jonathan-gramain (approved):

tmacro:
Upstream has removed the python script that generates the random seeds during initial startup, and replaced it with a java implementation. I've updated `10-generate-hash-seeds` to use the new one.
I've also taken the opportunity to remove `40-generate-secrets` since we use a static token and I suspect a few errors seen in the field to be cause by some flakiness here.
Finally I've added `ENV S6_BEHAVIOUR_IF_STAGE2_FAILS 2` which will cause s6 to exit on any failure during `cont-init` rather than continuing with startup.

tmacro (approved):

rahulreddy:
> Upstream has removed the python script that generates the random seeds during initial startup, and replaced it with a java implementation. I've updated `10-generate-hash-seeds` to use the new one.
> I've also taken the opportunity to remove `40-generate-secrets` since we use a static token and I suspect a few errors seen in the field to be cause by some flakiness here.
> Finally I've added `ENV S6_BEHAVIOUR_IF_STAGE2_FAILS 2` which will cause s6 to exit on any failure during `cont-init` rather than continuing with startup.

How do you usually track/figure out these changes?

PR #901, opened 2021-02-09
bf(S3C-3940): Prevent reporting of negative metrics

tmacro:


libV2/server/API/metrics/listMetrics.js
```diff
@@ -16,6 +16,14 @@ const metricResponseKeys = {
     service: 'serviceName',
 };
 
+function positiveOrZero(value) {
+    if (value < 0) {
+        return 0;
+    }
+
+    return value;
+}
```
miniscruff: Can use math max https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max

tmacro: fancy

libV2/server/API/metrics/getStorage.js
```diff
@@ -54,7 +54,7 @@ async function getStorage(ctx, params) {
 
     ctx.results.statusCode = 200;
     ctx.results.body = {
-        storageUtilized,
+        storageUtilized: storageUtilized > 0 ? storageUtilized : 0,
```
miniscruff: math max works here too

miniscruff (approved):
would like a test or two but looks good to me

anurag4DSB (approved):

libV2/server/API/metrics/getStorage.js
```diff
@@ -54,7 +54,7 @@ async function getStorage(ctx, params) {
 
     ctx.results.statusCode = 200;
     ctx.results.body = {
-        storageUtilized,
+        storageUtilized: Math.max(storageUtilized, 0),
```
anurag4DSB, line 57: Can it go in negative as well?

jonathan-gramain (approved):

libV2/server/API/metrics/getStorage.js
tmacro: Technically all UTAPI metrics can go negative as it is, at its core, just a system of summing numbers that you give it. This means that if the `+` metrics are missed but the `-` ones aren't you get negatives.

PR #908, opened 2021-02-11
Improvement/s3 c 3971 reduce event footprint

tmacro:
Includes an update to the latest arsenal to make the cloudserver update less painful

### DIFF
--- a/libV2/tasks/IngestShard.js
+++ b/libV2/tasks/IngestShard.js
@@ -19,6 +19,15 @@ class IngestShardTask extends BaseTask {
         super(options);
         this._defaultSchedule = config.ingestionSchedule;
         this._defaultLag = config.ingestionLagSeconds;
+        this._stripEventUUID = options.stripEventUUID !== undefined ? options.stripEventUUID : true;
+    }
+
+    _hydrateEvent(data) {
+        const event = JSON.parse(data);
+        if (this._stripEventUUID) {
+            delete event.uuid;
+        }
+        return new UtapiMetric(event);
     }
 
     async _execute(timestamp) {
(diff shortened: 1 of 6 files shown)


libV2/tasks/IngestShard.js
```diff
@@ -47,16 +56,16 @@ class IngestShardTask extends BaseTask {
                         if (shardAge < checkpointLagMicroseconds) {
                             metricClass = 'utapi.event';
                             records = metrics
-                                .map(m => new UtapiMetric(JSON.parse(m)));
+                                .map(m => this._hydrateEvent(m));
```
miniscruff, line 59: is it possible to remove the extra function wrap?
```suggestion
                                .map(this._hydrateEvent);
```

tests/functional/v2/task/testIngestShards.js
```diff
@@ -123,5 +126,23 @@ describe('Test IngestShards', function () {
         );
         assertResults(events, series);
     });
+
+    it('should strip the event uuid during ingestion', async () => {
+        const start = shardFromTimestamp(getTs(-120));
+        const stop = start + 9000000;
+        const events = generateFakeEvents(start, stop, 100);
+
+        await Promise.all(events.map(e => cacheClient.pushMetric(e)));
```
miniscruff, line 135: might also be able to remove the function wrap here too
```suggestion
        await Promise.all(events.map(cacheClient.pushMetric));
```

miniscruff (approved):

libV2/tasks/IngestShard.js
tmacro: You can do something like `this._hydrateEvent.bind(this)`, but it's just doing the same thing with a different syntax. Root problem is that js doesn't carry around a reference to the correct `this`, so in `_hydrateEvent` on L#27 without the wrap/bind `this` would not be pointing to an instance of `IngestShardTask`.

tests/functional/v2/task/testIngestShards.js
tmacro: See above

libV2/tasks/IngestShard.js
miniscruff: right... that dang `this`...

ilkescality (approved):

PR #928, opened 2021-02-22
bugfix: S3C-3620 don't raise exception on ioredis client error

jonathan-gramain:
When ioredis emits an error (e.g. connection issue), instead of
 unconditionally raising it in the RedisClient wrapper, only raise it
 if there is at least one listener for the 'error' event.
 
 Some users of RedisClient (v2 cache and reindex tasks) do not set a
 listener on error events, resulting in assertions being raised and
 process crash.

### DIFF
--- a/tests/functional/v2/cache/testRedisBackend.js
+++ b/tests/functional/v2/cache/testRedisBackend.js
@@ -9,8 +9,6 @@ const startTime = new Date().getTime();
 const endTime = startTime + 30000; // Add 30 seconds;
 const testValues = generateFakeEvents(startTime, endTime, 50);
 
-describe('test', async () => true);
-
 describe('Test cache redis backend', () => {
     let cache;
     let prefix;
--- a/tests/unit/v2/RedisClient.js
+++ b/tests/unit/v2/RedisClient.js
@@ -1,0 +1,21 @@
+const RedisClient = require('../../../libV2/redis');
+
+describe('Test RedisClient', () => {
+    let client;
+
+    beforeEach(() => {
+        client = new RedisClient({});
+        client.connect();
+    });
+
+    afterEach(() => client.disconnect());
+
+    it('should not raise exception if redis backend emits error without a listener', () => {
+        client._redis.emit('error', new Error('OOPS'));
+    });
+
+    it('should be able to listen to redis backend errors', done => {
+        client.on('error', () => done());
+        client._redis.emit('error', new Error('OOPS'));
+    });
+});
--- a/libV2/redis.js
+++ b/libV2/redis.js
@@ -103,7 +103,9 @@ class RedisClient extends EventEmitter {
 
     _onError(error) {
         moduleLogger.error('error connecting to redis', { error });
-        this.emit('error', error);
+        if (this.listenerCount('error') > 0) {
+            this.emit('error', error);
+        }
     }
 
     _createCommandTimeout() {


tmacro (approved):

tests/functional/v2/cache/testRedisBackend.js
```diff
@@ -9,8 +9,6 @@ const startTime = new Date().getTime();
 const endTime = startTime + 30000; // Add 30 seconds;
 const testValues = generateFakeEvents(startTime, endTime, 50);
 
-describe('test', async () => true);
```
miniscruff, line 12: :laughing:

tests/unit/v2/RedisClient.js
```diff
@@ -0,0 +1,21 @@
...
+
+describe('Test RedisClient', () => {
+    let client;
+
+    beforeEach(() => {
+        client = new RedisClient({});
+        client.connect();
+    });
+
+    afterEach(() => client.disconnect());
```
miniscruff, line 11: Does this work? I am not sure if this results in some weird race condition.
```suggestion
    afterEach(client.disconnect);
```

tests/unit/v2/RedisClient.js
```diff
@@ -0,0 +1,21 @@
...
+
+    it('should not raise exception if redis backend emits error', () => {
+        client._redis.emit('error', new Error('OOPS'));
+    });
+
+    it('should be able to listen to redis backend errors', done => {
+        client.on('error', () => done());
```
miniscruff, line 18:
```suggestion
        client.on('error', done);
```

tests/unit/v2/RedisClient.js
```diff
@@ -0,0 +1,21 @@
...
+    let client;
+
+    beforeEach(() => {
+        client = new RedisClient({});
+        client.connect();
+    });
+
+    afterEach(() => client.disconnect());
+
+    it('should not raise exception if redis backend emits error', () => {
```
miniscruff: could add an extra point to the it saying something like "if redis backend emits error without a listener"? Or equivalent as it wasn't quite clear to me at first.

tests/unit/v2/RedisClient.js
jonathan-gramain: This would pass on the error argument to the `done` callback and fail the test, so we need to pass a specific function.

tests/unit/v2/RedisClient.js
jonathan-gramain: Not sure I get what you mean. I guess we could but we would need to bind to the object instance by appending `.bind(client)`. I copied this logic from another test suite without thinking too much about it. It seems to handle both callback and async syntax, so the way it is I think works by returning a promise to `afterEach`.

tests/unit/v2/RedisClient.js
jonathan-gramain: done

tests/unit/v2/RedisClient.js
miniscruff: ooh good call forget about that.

tests/unit/v2/RedisClient.js
miniscruff: ok yea, I still gotta get used to bind.

miniscruff (approved):

PR #961, opened 2021-03-02
bf(S3C-4057): Handle timestamp conflicts during ingestion

tmacro:


libV2/config/index.js
```diff
@@ -271,7 +271,7 @@ class Config {
             warp10Conf.hosts = [{
                 host: _loadFromEnv('WARP10_HOST', config.warp10.host),
                 port: _loadFromEnv('WARP10_PORT', config.warp10.port, _typeCasts.int),
-                nodeId: parsedConfig.nodeId,
+                nodeId: _loadFromEnv('WARP10_NODE_ID', config.warp10.nodeId),
```
vrancurel, line 274: will it be the node host ?

libV2/utils/timestamp.js
```diff
@@ -19,16 +19,16 @@ class InterpolatedClock {
...
 
-    getTs() {
-        const ts = Date.now();
+    getTs(timestamp) {
+        const ts = timestamp !== undefined ? timestamp : Date.now();
         if (ts === this._now) {
             // If this is the same millisecond as the last call
             this._step += 1;
-            return ts * 1000 + (this._step - 1);
+            return convertTimestamp(ts) + (this._step - 1);
```
vrancurel, line 27: what if the next one also collide ?

tests/functional/v2/task/testIngestShards.js
```diff
@@ -144,5 +144,46 @@ describe('Test IngestShards', function () {
         );
         series[0].values.forEach(val => assert.strictEqual(val.id, undefined));
     });
+
+    it('should increment microseconds for duplicate timestamps', async () => {
```
vrancurel, line 148: Do we test the case where there are >2 events conflicting on the same ts ?

libV2/utils/timestamp.js
tmacro: `this._step` is incremented each time.

libV2/config/index.js
tmacro: `config.warp10.nodeId`  will be the host warp 10 is on (used for `node` label). The top level `nodeId` is the host the utapi process is running on (used for `origin` label)

tests/functional/v2/task/testIngestShards.js
tmacro: I'll add one

vrancurel (approved):

jonathan-gramain (approved):

PR #968, opened 2021-03-03
bf(S3C-4049): Call delete in slices rather than once

tmacro:


alexanderchan-scality (approved):

libV2/tasks/DiskUsage.js
```diff
@@ -84,16 +86,19 @@ class MonitorDiskUsage extends BaseTask {
...
-        await this.withWarp10(async warp10 =>
-            warp10.delete({
-                className: '~.*',
-                start: oldestTimestamp - 1,
-                end: endTimestamp,
-            }));
+        await async.eachSeries(
+            sliceTimeRange(oldestTimestamp - 1, endTimestamp, expirationChunkDuration),
+            async ([start, end]) => {
+                moduleLogger.info('deleting metrics',
```
jonathan-gramain, line 92: It could be better to log `start` and `end` instead, or in addition to oldest/end timestamps

libV2/utils/timestamp.js
```diff
@@ -42,8 +42,45 @@ function now() {
...
+ *
+ * @param {Number} start
+ * @param {Number} end
+ * @param {Number} step
+ */
+
+function* sliceTimeRange(start, end, step) {
+    let spos = start;
+    let epos = start + step - 1;
+    while (true) {
```
jonathan-gramain: A minor improvement/clarification of this loop could be:

```javascript
while (epos < end) {
    yield [spos, epos];
    spos += step;
    epos += step;
}
yield [spos, end];
```

naren-rajendran (approved):

libV2/tasks/DiskUsage.js
tmacro: ah yeah thats what I meant to do

libV2/utils/timestamp.js
tmacro: I actually had this to start with and changed it because I though there was an edge case it didn't handle. Upon a second look that edge case isn't possible :D

libV2/tasks/DiskUsage.js
jonathan-gramain: don't forget to change it :wink:

jonathan-gramain (approved):

PR #1013, opened 2021-03-15
Bugfix/s3 c 4139 allow only start timestamp in req

tmacro:


naren-rajendran (approved):

tests/functional/v2/server/testListMetrics.js
```diff
@@ -30,10 +30,15 @@ const emptyOperationsResponse = Object.values(operationToResponse)
...
     const body = {
-        timeRange: [start, end],
         [level]: resources,
     };
 
+    if (end !== undefined) {
+        body.timeRange = [start, end];
+    } else {
+        body.timeRange = [start];
+    }
```
miniscruff, line 40: I think `[start, end]` works in both cases as end would be undefined.

miniscruff (approved):

alexanderchan-scality (approved):

tests/functional/v2/server/testListMetrics.js
tmacro: it serializes to `[0, null]` for some reason if you leave it in

PR #1076, opened 2021-05-26
bf(S3C-4429): null sizeD calculated during reindex

tmacro:
Add some additional error handling around the `_fetchCurrentMetrics` method to prevent bad db values from causing a crash.
Swap from `getMetrics` to `getMetricsAt` for some free perf, as we only need the metrics at a single timestamp.


jonathan-gramain (approved):

ilkescality (approved):

mrepop (approved):
LGTM

PR #1082, opened 2021-06-23
bf(S3C-4550): Do not take previous reindex diffs into account when calculating its own

tmacro:
As only the latest reindex diff is taken into account by `getMetricsAt` the diffs calculated can begin "flapping" as it over/under corrects. See ticket for a detailed explanation.

### DIFF
--- a/libV2/tasks/Reindex.js
+++ b/libV2/tasks/Reindex.js
@@ -69,6 +69,8 @@ class ReindexTask extends BaseTask {
                     labels: {
                         [level]: resource,
                     },
+                    // eslint-disable-next-line camelcase
+                    no_reindex: true,
                 },
                 macro: 'utapi/getMetricsAt',
             };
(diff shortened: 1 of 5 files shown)


libV2/tasks/Reindex.js
```diff
@@ -69,6 +69,8 @@ class ReindexTask extends BaseTask {
                     labels: {
                         [level]: resource,
                     },
+                    // eslint-disable-next-line camelcase
+                    no_reindex: false,
```
jonathan-gramain: is `false` the default value? Wondering since in this PR I don't see any place where it would change the current behavior, since I expect `<% $no_reindex NOT %>` to always be true then. I may totally be wrong, just wondering.

Also it may (or not) make more sense to reverse the logic with a `reindex` flag (or named differently) to avoid double negations.

saitharunthodupunuru (approved):

libV2/tasks/Reindex.js
tmacro: this should be `true`, leftover from testing. The reason for the inverted logic: warpscript returns a `NULL` value for keys that aren't present like javascript, which allows us to introduce this flag without upstream callers being aware as it needs to be explicitly set to `true`.

jonathan-gramain: OK makes sense, the `false` value was confusing me :)

jonathan-gramain (approved):

PR #1092, opened 2021-07-22
(draft) exporters for default nodejs and redis metrics

watsaqat:
This provides metrics for NodeJs and Redis which Prometheus can scrape. To scrape Redis, an [exporter](https://github.com/oliver006/redis_exporter) is used as a service which also provides metrics about itself (scrape durations, memory usage, etc). Additionally, warp10 metrics and custom metrics for Utapi server will be added soon. 

Currently, the available metrics are:
    NodeJs:                 https://github.com/siimon/prom-client/tree/master/lib/metrics
    Redis:                    https://redis.io/commands/info
    Redis-Exporter:    https://github.com/oliver006/redis_exporter/blob/master/exporter/exporter.go (starting at line 322)
    
The functionality in this draft only targets Utapi V2.

### DIFF
--- a/libV2/server/API/internal/prometheusMetrics.js
+++ b/libV2/server/API/internal/prometheusMetrics.js
@@ -1,0 +1,15 @@
+const { collectDefaultMetrics, register } = require('prom-client');
+
+collectDefaultMetrics({
+    timeout: 10000,
+    gcDurationBuckets: [0.001, 0.01, 0.1, 1, 2, 5],
+    prefix: 'Utapi_v2_',
+});
+
+async function prometheusMetrics(ctx) {
+    // eslint-disable-next-line no-param-reassign
+    ctx.results.statusCode = 200;
+    ctx.results.body = await register.metrics();
+}
+
+module.exports = prometheusMetrics;
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,45 @@
+version: '3.8'
+
+x-models:
+  warp10: &warp10
+    build:
+      context: .
+      dockerfile: ./images/warp10/Dockerfile
+    volumes: [ $PWD/warpscript:/usr/local/share/warpscript ]
+
+  warp10_env: &warp10_env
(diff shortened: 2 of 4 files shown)


libV2/server/API/internal/prometheusMetrics.js
```diff
@@ -0,0 +1,15 @@
+const { collectDefaultMetrics, register } = require('prom-client');
+
+collectDefaultMetrics({
+    timeout: 10000,
+    gcDurationBuckets: [0.001, 0.01, 0.1, 1, 2, 5],
+    prefix: 'Utapi_v2_',
```
jonathan-gramain, line 6: The prefix looks fine, just might be more consistent with other prometheus metric names if all lowercase

jonathan-gramain (commented):
Looks like a good start!

Just about the redis exporter, no problem to publish the `docker-compose,yml` somewhere if it helps setting up a dev environment, for open source etc. (here or somewhere else, not sure what is the best place), but there will need to be changes in Federation to have the exporter setup at deployment time with redis.

Not sure what is the best way to do this, one idea is to build our own Dockerfile based on the redis image currently deployed, to add the exporter (e.g. by downloading it and adding it to the image, setting it up to monitor the local redis instance, and starting it up when the container is started).

libV2/server/API/internal/prometheusMetrics.js
jonathan-gramain: Now that I think about it, the default metrics may be exported without a separate prefix. Indeed Prometheus allows to fetch metrics by label, and those metrics could (should) be labeled by service, hence they could be queried targeting utapi only by providing the proper label selector (say, `{service="utapi"}`). This allows querying those metrics more generally as well.

jonathan-gramain: I'm suggesting this since the metrics are applicable to any node.js process, not just utapi. For utapi-specific metrics, e.g. if it was the number of accounts monitored, it would make sense to prefix them.

PR #1138, opened 2021-09-28
ft(UTAPI-24): Limit user credentials

tmacro:


libV2/vault/index.js
```diff
@@ -73,42 +73,87 @@ async function authorizeAccountAccessKey(authInfo, level, resources, log) {
...
+
+    log.trace('Authorizing IAM user', { resources });
+
+    // Get the parent account id from the user's arn
+    const parentAccountId = authInfo.getArn().split(':')[4];
+
+    // All users require an attached policy to query metrics
+    // Additional filtering is performed here to limit access to the user's account
+    switch (level) {
+    // User keys can only query metrics their own account metrics
```
jonathan-gramain, line 91:
```suggestion
    // User keys can only query their own account metrics
```

mrepop (approved):

libV2/vault/client.js
```diff
@@ -76,6 +77,20 @@ class Vault extends auth.Vault {
...
 
+                    // Only IAM users will return authorizedResources
+                    let authorizedResources = resources;
+                    if (authRes) {
+                        authorizedResources = authRes
+                            .filter(resource => resource.isAllowed)
+                            .map(resource => {
+                                // resource.arn should be of format:
+                                // arn:scality:utapi:::resourcetype/resource
+                                assert(typeof resource.arn === 'string');
```
jonathan-gramain, line 89:
```suggestion
                                assert.strictEqual(typeof resource.arn, 'string');
```

libV2/vault/index.js
```diff
@@ -13,11 +12,12 @@ async function translateResourceIds(level, resources, log) {
     return resources.map(resource => ({ resource, id: resource }));
 }
 
-
 async function authorizeAccountAccessKey(authInfo, level, resources, log) {
     let authed = false;
     let authedRes = [];
 
+    log.trace('Authorizing account', { resources });
```
jonathan-gramain, line 19: Could be useful to log `authInfo` too (possibly pruned of its security-sensitive content if any) and `level`

libV2/vault/index.js
```diff
@@ -73,42 +73,87 @@ async function authorizeAccountAccessKey(authInfo, level, resources, log) {
...
-            return authed;
-        }, [],
-    );
-
-    return [authedRes.length !== 0, authorizedResources];
+async function authorizeUserAccessKey(authInfo, level, resources, log) {
+    let authed = false;
+    let authedRes = [];
+
+    log.trace('Authorizing IAM user', { resources });
```
jonathan-gramain, line 83: same for logging `authInfo`

libV2/vault/index.js
```diff
@@ -73,42 +73,87 @@ async function authorizeAccountAccessKey(authInfo, level, resources, log) {
...
+    // User keys can only query metrics their own account metrics
+    // So we can short circuit the auth to ->
+    // Did they request their account? Then authorize ONLY their account
+    case 'accounts': {
+        authed = resources.some(r => r === parentAccountId);
+        authedRes = authed ? [{ resource: parentAccountId, id: authInfo.getCanonicalID() }] : [];
+        break;
+    }
+
+    // Users can query other user's metrics if they are under the same account
```
jonathan-gramain, line 100: It looks surprising to me, at first I would think that a user may not be able to access other users' metrics, since he may not be able to access other users' buckets. But I'm not familiar with the specs here so just wondering.

tmacro: By default users have access to nothing. You need to attach a user policy to grant using the usual `<type>/<resource>` to give access. This check is so that a policy with `users/*` does not allow access to different accounts.

jonathan-gramain (approved):

PR #1141, opened 2021-09-30
ft(UTAPI-26): Add authorization for service user

tmacro:


mrepop (approved):
Should avoid hard coding the service level admin factors into the test suites and pull them from the environment setup if possible.

jonathan-gramain (approved):

PR #1142, opened 2021-10-01
bf(UTAPI-27): Fail bucketd request after repeated errors

tmacro:


lib/reindex/s3_bucketd.py
```diff
@@ -52,18 +53,24 @@ def inner(*args, **kwargs):
...
+        super().__init__('Max retries reached for request to %s'%url)
+
 class BucketDClient:
 
     '''Performs Listing calls against bucketd'''
     __url_format = '{addr}/default/bucket/{bucket}'
     __headers = {"x-scal-request-uids": "utapi-reindex-list-buckets"}
 
-    def __init__(self, bucketd_addr=None):
+    def __init__(self, bucketd_addr=None, max_retries=5):
```
jonathan-gramain: It could be better to stay consistent with the script default and set 2 retries by default here, or even define this number of retries in a script global constant.

lib/reindex/s3_bucketd.py
```diff
@@ -108,6 +117,9 @@ def _list_bucket(self, bucket, **kwargs):
                 _log.error('Invalid listing response body! bucket:%s params:%s'%(
                     bucket, ', '.join('%s=%s'%p for p in params.items())))
                 continue
+            except MaxRetriesReached as e:
```
jonathan-gramain:
```suggestion
            except MaxRetriesReached:
```

lib/reindex/s3_bucketd.py
```diff
@@ -337,16 +349,26 @@ def log_report(resource, name, obj_count, total_size):
...
+            jobs = { executor.submit(index_bucket, bucket_client, b): b for b in batch }
+            for job in futures.as_completed(jobs.keys()):
+                try:
+                    total = job.result() # Summed bucket and shadowbucket totals
+                except MaxRetriesReached:
+                    _bucket = jobs[job]
+                    _log.error('Failed to list bucket %s. Removing from results.'%_bucket.name)
+                    # Add the bucket to observed_buckets anyway to avoid clearing existing metrics
+                    observed_buckets.add(_bucket.name)
+                    # If we can not list one of an account's buckets we can not update it's total
```
jonathan-gramain:
```suggestion
                    # If we can not list one of an account's buckets we can not update its total
```

jonathan-gramain (approved):

mrepop (approved):

PR #1147, opened 2021-10-15
ft(UTAPI-33): Add ensureServiceUser script

tmacro:


bin/ensureServiceUser
```diff
@@ -0,0 +1,276 @@
...
+    ].map(h => new h(serviceName, client, log));
+}
+
+function apply(client, serviceName, log, cb) {
+    const handlers = buildServiceUserHandlers(serviceName, client, log);
+
+    async.waterfall([
+        done => collectResourcesFromHandlers(handlers, done),
+        ...handlers.map(h => h.applyWaterfall.bind(h)),
+        (values, done) => done(null, values.accessKey),
```
jonathan-gramain, line 222: Do we return `values.accessKey`on purpose here? Since it looks like `apply` is called for every resource type (not just `AccessKeyHandler`)

bin/ensureServiceUser
```diff
@@ -0,0 +1,276 @@
...
+        log.info('success', { data });
+        process.exit();
+    });
+}
+
+program.version(version);
+
+[
+    {
+        name: 'apply <service-name>',
```
jonathan-gramain, line 255: What does `apply` mean in this context? Could it be called `create` instead if it is used to create resources associated to the service user?

jonathan-gramain (approved):

saitharunthodupunuru (approved):

bin/ensureServiceUser
tmacro: This is mostly a copy/paste job from https://github.com/scality/backbeat/blob/development/8.3/bin/ensureServiceUser with the applied policy changed.
I believe its broken out so that the message print on L#246 contains it.

bin/ensureServiceUser
tmacro: see above

PR #1174, opened 2021-11-11
ft(UTAPI-43): Add allow/deny filter for events

tmacro:


libV2/utils/filter.js
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Constructs a function meant for filtering Objects by the value of a key
+ * Returned function returns a boolean with false meaning the object was present
+ * in the filter allowing the function to be passed directly to Array.filter etc.
+ *
+ * @param {string} key - Object key to inspect
+ * @param {Object} filter
+ * @param {Set} [filter.allow] - Set containing keys to include
+ * @param {Set} [filter.deny] - Set containing keys to not include
+ * @returns {bool}
```
watsaqat: returns a function instead of bool

watsaqat (approved):

naren-rajendran (approved):

libV2/config/index.js
```diff
@@ -225,6 +227,28 @@ class Config {
...
     }
 
+    static _parseResourceFilters(config) {
+        const resourceFilters = {};
+
+        allowedFilterFields.forEach(
+            field => allowedFilterStates.forEach(
+                state => {
+                    const configResources = (config[state] && config[state][field]) || null;
+                    const envVar = `FILTER_${field.toUpperCase()}_${state.toUpperCase()}`;
```
jonathan-gramain, line 237:
```suggestion
                    const envVar = `UTAPI_FILTER_${field.toUpperCase()}_${state.toUpperCase()}`;
```

libV2/tasks/Reindex.js
```diff
@@ -145,6 +155,11 @@ class ReindexTask extends BaseTask {
         const accountTotals = {};
         const ignoredAccounts = new Set();
         await async.eachLimit(this.targetBuckets, 5, async bucket => {
+            if (!this._shouldReindex({ bucket: bucket.name, account: bucket.account })) {
+                logger.debug('skipping excluded bucket.', { bucket: bucket.name, account: bucket.account });
```
jonathan-gramain:
```suggestion
                logger.debug('skipping excluded bucket', { bucket: bucket.name, account: bucket.account });
```

tests/unit/v2/utils/testFilter.js
```diff
@@ -0,0 +1,120 @@
...
+        key2: { allow: new Set(['allow']) },
+        msg: 'test chain with multiple allow filters',
+    },
+    {
+        key1: { allow: new Set(['allow']) },
+        key2: { deny: new Set(['deny']) },
+        msg: 'test chain with multiple deny filters',
+    },
+    {
+        key1: { deny: new Set(['deny']) },
```
jonathan-gramain:
```suggestion
        key1: { allow: new Set(['allow']) },
```

tests/unit/v2/utils/testFilter.js
```diff
@@ -0,0 +1,120 @@
...
+
+
+const chainTestCases = [
+    {
+        key1: { allow: new Set(['allow']) },
+        key2: { allow: new Set(['allow']) },
+        msg: 'test chain with multiple allow filters',
+    },
+    {
+        key1: { allow: new Set(['allow']) },
```
jonathan-gramain:
```suggestion
        key1: { deny: new Set(['deny']) },
```

libV2/client/index.js
```diff
@@ -84,6 +89,11 @@ class UtapiClient {
         this._drainCanSchedule = true;
         this._drainDelay = (config && config.drainDelay) || 30000;
         this._suppressedEventFields = (config && config.suppressedEventFields) || null;
+        const eventFilters = (config && config.filter) || {};
+        this._shouldPushMetric = buildFilterChain(eventFilters);
```
jonathan-gramain, line 93: I'm not sure if building a filter chain like done isn't overengineered, instead of a simpler existence check of keys in two sets, allowed and denied, pre-filled from the configuration (or just `denied` since `allowed` may not be needed per the requirements)

Something like

```
for each key passed to the filter function
    if there is an `allow` set
        if the key exists in the set, return "allow"
        else return "deny"
    else if there is a "deny" set
        if the key exists in the set, return "deny"
        else return "allow"

File: libV2/utils/filter.js [hunk 7]
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+function filterObject(key, { allow, deny }) {
+    if (allow && deny) {
+        throw new Error('You can not define both an allow and a deny list.');
+    }
+    if (!allow && !deny) {
+        throw new Error('You must define either an allow or a deny list.');
+    }
+    if (allow) {
```
jonathan-gramain, line 24: Is it worth having an allow set, since AFAIK it was not part of the requirements? Ok to keep it but we need to be wary of the added complexity.

libV2/config/index.js
tmacro: The `UTAPI` bit is added in `_loadFromEnv`

libV2/client/index.js
tmacro: While I agree that the chain is more complex that a loop, I think that avoiding the overhead of the loop and checks at runtime is worth it given that this will be called on most API calls.

libV2/utils/filter.js
tmacro: I think, from a technical perspective, that implementing `allow` at the same time as `deny` makes sense as it allows us to design an initial solution that supports both rather than needing rework if the feature is eventually asked for. In my opinion the small amount of extra code and complexity needed to support `allow` is worth it.

libV2/client/index.js
jonathan-gramain: Correct me if I'm wrong, what I understand that we are mostly saving is the `if` condition on whether we should check on the allow or deny set since it is memoized by creating an ad-hoc function on either of the sets. We are also indeed not using a loop over the keys, but replacing it by nested calls to the ad-hoc function, which I'm not convinced is more efficient than a loop since function calls also have a cost in themselves.

Overall I feel we are trying to save on crumbs at the expense of complexity and we may not even achieve the intended savings, hence I believe it is best not to try to optimize at this point, since `Set` structures should be efficient enough for the need and I believe should not incur a measurable overhead (but maybe I'm wrong, just my feeling).

jonathan-gramain: We are actually not saving on the lookups in the `Set`, but my point is that with a simple loop we should get the expected performance anyways, we can always revisit this if we observe a noticeable impact.

One random idea if that turns out to be the case that could be a more efficient optimization is using a bloom filter on the set of keys that we want to deny, so that we can check quickly if none of them is in the set with high probability.

libV2/utils/filter.js
jonathan-gramain: OK, good for me it makes sense.

jonathan-gramain (approved):

PR #1190, opened 2021-11-22
impr(S3C-4767): skip objects without a content-length during reindex

tmacro:


libV2/tasks/Reindex.js
```diff
@@ -51,6 +51,13 @@ class ReindexTask extends BaseTask {
                 // eslint-disable-next-line no-continue
                 continue;
             }
+
+            if (!Number.isInteger(obj.value['content-length'])) {
+                logger.debug('object missing content-length, not including in count');
```
jonathan-gramain, line 56: Actually, a missing content-length refers to a 0 byte object, which I think should still be included in the `count` right?

Instead of this check, what about changing line 62 to:

```
size += obj.value['content-length'] || 0;
```

jonathan-gramain: I believe those are all 0 byte objects, maybe there are other objects that we do not want to count, not sure but I don't think so.

tmacro: zero byte objects have a content-length of `0`

jonathan-gramain: Ah ok, so then it looks good as is.

jonathan-gramain (approved):

watsaqat (approved):

PR #1209, opened 2022-01-06
impr(UTAPI-58): Limit maximum size of snapshots

tmacro:


warpscript/utapi/createSnapshot.mc2
```diff
@@ -40,15 +40,54 @@
...
+            STOP
+          %> IFT
+        %>  IFT
+        <% $oldestTimestamp $max_snapshot_size + $endTimestamp < %>
+        <%
+          $oldestTimestamp $max_snapshot_size + 'endTimestamp' STORE
+        %> IFT
+      %> 
+      <%
+        $masterSnapshotTimestamp $max_snapshot_size + 'endTimestamp' STORE
```
jonathan-gramain: Not sure if I get this right, if here we are in the "else" clause of this check at https://github.com/scality/utapi/pull/1209/files#diff-0547afee6eaac48bd3f26855e5b9adbcb9bd4e056ba6d04ef792027da0a93b0aR57:

```
<% $endTimestamp $masterSnapshotTimestamp - $max_snapshot_size > %>
```

Wouldn't this mean we are requesting a range less than 6 hours since the last master snapshot, hence keep the end timestamp as requested, instead of using `$max_snapshot_size` and storing it in `endTimestamp`?

tmacro: The nesting is a bit confusing, but this is the else for `<% $masterSnapshotTimestamp 0 == %>`
Meaning we are in the "We have a previous snapshot" case

jonathan-gramain: ah indeed

jonathan-gramain (approved):

watsaqat (approved):

PR #1212, opened 2022-01-25
UTAPI-59 upgrade to Node 16

nicolas2bert:


nicolas2bert:
ping

miniscruff (approved):

nicolas2bert:
@bert-e reset

lib/UtapiReindex.js
```diff
@@ -106,7 +106,7 @@ class UtapiReindex {
     _runScript(path, done) {
         const flags = this._buildFlags();
         this._requestLogger.debug(`launching subprocess ${path} with flags: ${flags}`);
-        const process = childProcess.spawn('python3.4', [
+        const process = childProcess.spawn('python3.7', [
```
nicolas2bert: @tmacro is it ok to use `python3.7` here?
Now that we upgrade to Node 16, the base image uses Debian 10 that includes Python version 3.7

nicolas2bert:
@bert-e wait

lib/UtapiReindex.js
tmacro: It should be, I regularly use 3.10 when I develop the script

tmacro (approved):

PR #1235, opened 2022-05-04
ft(UTAPI-67): Add metrics framework to BaseTask

tmacro:


docker-compose.yaml
```diff
@@ -19,31 +19,31 @@ x-models:
...
+
   redis: &redis
       build:
         context: .
         dockerfile: ./images/redis/Dockerfile
 
 services:
   redis-0:
-    image: redis:5
+    image: redis:6
```
francoisferrand, line 30: seems like you are smuggling a redis upgrade, would be worth moving this to a separate commit and mentioning it (either that is has been upgraded, or that it is a requirement to use at least version 6)...

libV2/tasks/BaseTask.js
```diff
@@ -39,6 +46,70 @@ class BaseTask extends Process {
...
+    }
+
+    _registerDefaultMetricHandlers() {
+        const taskName = this.constructor.name;
+
+        // Get the name of our subclass in snake case format eg BaseClass => _base_class
+        const taskNameSnake = taskName.replace(/[A-Z]/g, letter => `_${letter.toLowerCase()}`);
+
+        const executionTime = new promClient.Gauge({
+            name: `utapi${taskNameSnake}_execution_seconds`,
```
francoisferrand: * `execution_seconds` seems a bit strange, maybe `duration_seconds` would be more self-describing?
* instead of a gauge, it may be better to use two counter: `total_duration` and `count`, which can be much more easily managed afterwards (aggregated, mean...); to simplify this, an histogram or summary may be used.

libV2/tasks/BaseTask.js
```diff
@@ -39,6 +46,70 @@ class BaseTask extends Process {
...
+            help: `Execution time of the ${taskName} task`,
+            labelNames: ['origin', 'containerName'],
+        });
+
+        const executionAttempts = new promClient.Counter({
+            name: `utapi${taskNameSnake}_attempts_total`,
```
francoisferrand, line 72: not sure what this metrics is : is it the number of time the task was executed, or some more specific metrics due some kind of smart scheduling (and thus a task may be de-scheduled/...)

The "attempt" here makes me think there is something more, but it is not clear what.
If it is just the number of executions, it may part of the historgram/summary above (see previous comment).

libV2/tasks/BaseTask.js
```diff
@@ -39,6 +46,70 @@ class BaseTask extends Process {
...
+            name: `utapi${taskNameSnake}_failures_total`,
+            help: `Number of failures executing the ${taskName} task`,
+            labelNames: ['origin', 'containerName'],
+        });
+
+        return {
+            executionTime,
+            executionAttempts,
+            executionFailures,
+        };
```
francoisferrand, line 87: i guess utapi is exposing the usage via some API : in that case, it should also include the "standard" http metrics (count, duration, size...), like vault and cloudserver:
* [http_requests_total](https://github.com/scality/cloudserver/blob/HEAD/lib/utilities/monitoringHandler.js#L39)
* [http_request_duration_seconds](https://github.com/scality/cloudserver/blob/HEAD/lib/utilities/monitoringHandler.js#L44)
* [http_active_requests](https://github.com/scality/cloudserver/blob/HEAD/lib/utilities/monitoringHandler.js#L50)

libV2/config/schema.js
```diff
@@ -115,7 +115,17 @@ const schema = Joi.object({
             return filterObj;
         }, {},
     )),
+    metrics: {
+        enabled: Joi.boolean(),
+        host: Joi.string(),
+        serverPort: Joi.number().port(),
```
francoisferrand: why do we need to use different metrics port?

a single metrics endpoint can be used to serve the metrics from all these endpoints; and it will serve only the metrics which are actually registered (e.g. if one instance runs only some tasks, it will return only that tasks' metrics)

having a single port to scrape will be slightly better in terms of (prometheus) load ; and simplify configuration/integration (esp. in Kubernetes, where we would need to declare extra port mappings to auto-detection)

francoisferrand: also, for consistency : the names used here (in config) should somehow match the names of the metrics (which are currently the name of the classes)

libV2/tasks/BaseTask.js
```diff
@@ -39,6 +46,70 @@ class BaseTask extends Process {
...
+        });
+
+        this._probeServer.addHandler(
+            DEFAULT_METRICS_ROUTE,
+            (res, log) => {
+                log.debug('metrics requested');
+                res.writeHead(200, {
+                    'Content-Type': promClient.register.contentType,
+                });
+                promClient.register.metrics().then(metrics => {
```
francoisferrand, line 108: the code is using async/await anyway, so you may as well use it here as well

```
res.end(await promClient.register.metrics());
```

libV2/tasks/BaseTask.js
tmacro: My reasoning behind using a  gauge to record each duration rather than a histogram is that these tasks run relatively infrequently (fastest being every 30 secs) and that the volume of metrics produced would be low enough to prefer the extra information. This would be especially handy with tasks like snapshots that only run every hour.

libV2/tasks/BaseTask.js
tmacro: Its just the number of executions. The reason I went with `attempts` was I didn't want to imply that it was successful,  though that may be thinking a bit to much into it.

libV2/tasks/BaseTask.js
tmacro: The http server metrics will need to be handled a bit differently since the server doesn't inherit from BaseTask

libV2/tasks/BaseTask.js
tmacro: This particular handler isn't async

docker-compose.yaml
tmacro: The upgrade landed in s3c last week, this docker-compose is just for development

libV2/config/schema.js
tmacro: Forgot to comment yesterday on this, the different ports are needed because the tasks don't run in the same process.

jonathan-gramain (approved):

watsaqat (approved):

PR #1239, opened 2022-05-06
ft(UTAPI-69): Add metrics for async tasks

tmacro:


libV2/tasks/DiskUsage.js
```diff
@@ -46,6 +47,64 @@ class MonitorDiskUsage extends BaseTask {
...
+            labelNames: ['origin', 'containerName'],
+        });
+
+        const diskUsage = new promClient.Gauge({
+            name: 'utapi_monitor_disk_usage_bytes',
+            help: 'Total bytes used by warp 10',
+            labelNames: ['origin', 'containerName'],
+        });
+
+        const hardLimitRatio = new promClient.Gauge({
```
jonathan-gramain, line 64: Instead of storing the ratio in Prometheus, it might be better to store the hard limit value itself, so likely it will be constant until it gets changed, to follow the philosophy of storing the raw values instead of pre-computed ones. The ratio can be simply computed with a query doing the division with the disk space used, and it allows other types of queries like how much space remains before hitting the hard limit. It might be useful also to know if/when the hard limit was changed, which will be obvious when storing the hard limit value itself.

libV2/tasks/IngestShard.js
```diff
@@ -28,6 +29,52 @@ class IngestShardTask extends BaseTask {
...
 
+    // eslint-disable-next-line class-methods-use-this
+    _registerMetricHandlers() {
+        const ingestedTotal = new promClient.Counter({
+            name: 'utapi_ingest_shard_task_ingest_total',
+            help: 'Number of metrics ingested',
+            labelNames: ['origin', 'containerName'],
+        });
+
+        const ingestedSlow = new promClient.Counter({
```
jonathan-gramain, line 40: In addition to this, we could store the sum of the ingestion lag (what is computed as `shardAge` if that makes sense), as a counter, this way we can easily query the average lag over a certain time range or other aggregating queries.

libV2/tasks/IngestShard.js
```diff
@@ -28,6 +29,74 @@ class IngestShardTask extends BaseTask {
...
+
+        const ingestedShards = new promClient.Counter({
+            name: 'utapi_ingest_shard_task_shard_ingest_total',
+            help: 'Number of slow metrics ingested',
+            labelNames: ['origin', 'containerName'],
+        });
+
+        const shardAgeTotal = new promClient.Counter({
+            name: 'utapi_ingest_shard_task_shard_age_total',
+            help: 'Number of slow metrics ingested',
```
jonathan-gramain:
```suggestion
            help: 'Aggregated age of shards',
```

libV2/tasks/IngestShard.js
```diff
@@ -28,6 +29,74 @@ class IngestShardTask extends BaseTask {
...
+            ingestedShards,
+            shardAgeTotal,
+        };
+    }
+
+    /**
+     * Metrics for IngestShardTask
+     * @typedef {Object} IngestShardMetrics
+     * @property {number} ingestedTotal - Number of events ingested
+     * @property {number} ingestedSlow - Number of slow events ingested
```
jonathan-gramain, line 70: Missing info about `ingestedShards` and `shardAgeTotal`

libV2/tasks/DiskUsage.js
```diff
@@ -46,6 +47,75 @@ class MonitorDiskUsage extends BaseTask {
...
+            hardLimitSetting,
+        };
+    }
+
+    /**
+     * Metrics for MonitorDiskUsage
+     * @typedef {Object} MonitorDiskUsageMetrics
+     * @property {boolean} isLocked - Indicates if writes have been disabled for the monitored warp10
+     * @property {number} diskUsage - Port to bind probe server to
+     * @property {number} hardLimitRatio - Port to bind probe server to
```
jonathan-gramain: missing info about `hardLimitSetting`

jonathan-gramain (approved):

watsaqat (approved):

PR #1241, opened 2022-05-10
ft(UTAPI-70): Add http server metrics

tmacro:


libV2/server/metrics.js
```diff
@@ -0,0 +1,19 @@
...
+    name: 'utapi_http_requests_total',
+    help: 'Total number of HTTP requests',
+    labelNames: ['action', 'code'],
+});
+
+const httpRequestDurationSeconds = new promClient.Histogram({
+    name: 'utapi_http_request_duration_seconds',
+    help: 'Duration of HTTP requests in seconds',
+    labelNames: ['action', 'code'],
+    buckets: [0.0001, 0.005, 0.015, 0.05, 0.1, 0.2, 0.3, 0.4, 0.5], // buckets for response time from 0.1ms to 500ms
```
jonathan-gramain: Some abnormal requests may take more than 500ms, maybe even 10 seconds or more, and it can be very useful to know when those occur, so we could probably extend the bucket range, just following the same kind of exponential intervals but across a larger range, say from 0.1ms to 60s.

jonathan-gramain (approved):

watsaqat (approved):

PR #1243, opened 2022-05-17
impr(UTAPI-66): Migrate to arsenal 7.10.24

tmacro:


libV2/server/middleware.js
```diff
@@ -119,7 +119,7 @@ async function authV4Middleware(request, response, params) {
     } catch (error) {
         request.logger.error('error during authentication', { error });
         // rethrow any access denied errors
-        if (error.AccessDenied) {
+        if (error && (error.is || {}).AccessDenied) {
```
jonathan-gramain: no big deal, but it could be better to be consistent in how we check errors regarding the new style, i.e. choosing a coding style, like compared to https://github.com/scality/utapi/pull/1243/files#diff-664226905101de232ca7df5e0172bc90bbf7cf1e3b26a64e320b37ee0713dfbdR106

Also here I guess the `error &&` is not needed since we are in the catch block

tests/unit/testMetrics.js
```diff
@@ -53,13 +53,12 @@ function assertMetrics(schemaKey, metricName, props, isNegativeValue, done) {
...
         (err, res) => {
             if (isNegativeValue) {
-                assert.deepStrictEqual(
-                    err,
-                    errors.InternalError.customizeDescription(
-                        'Utapi is in a transient state for this time period as '
-                            + 'metrics are being collected. Please try again in a few '
-                            + 'minutes.',
-                    ),
+                assert(err.is[errors.InternalError.type]);
```
jonathan-gramain: Does this work instead?

```suggestion
                assert(err.is.InternalError);
```

jonathan-gramain (approved):

naren-rajendran (approved):

PR #1249, opened 2022-05-26
ft(UTAPI-75): Add metrics for latest checkpoint and snapshot

tmacro:
Also fixes UTAPI-22 so that the latest checkpoint dashboards work correctly

### DIFF
--- a/libV2/tasks/CreateCheckpoint.js
+++ b/libV2/tasks/CreateCheckpoint.js
@@ -29,8 +29,26 @@ class CreateCheckpoint extends BaseTask {
             labelNames: ['origin', 'containerName'],
         });
 
+        const getLastCheckpoint = this._getLastCheckpoint.bind(this);
+        const lastCheckpoint = new promClient.Gauge({
+            name: 'utapi_create_checkpoint_last_checkpoint_seconds',
+            help: 'Timestamp of the last successfully created checkpoint',
+            labelNames: ['origin', 'containerName'],
+            async collect() {
+                try {
+                    const timestamp = await getLastCheckpoint();
+                    if (timestamp !== null) {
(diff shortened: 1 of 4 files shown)


libV2/tasks/CreateCheckpoint.js
```diff
@@ -55,6 +73,25 @@ class CreateCheckpoint extends BaseTask {
...
+        const resp = await this.withWarp10(async warp10 => warp10.fetch({
+            className: 'utapi.checkpoint.master',
+            labels: {
+                node: warp10.nodeId,
+            },
+            start: 'now',
+            stop: -1,
+        }));
+
+        if (resp.result && (resp.result.length === 0 || resp.result[0] === '' || resp.result[0] === '[]')) {
```
jonathan-gramain: I believe we should also return `null` if `resp.result` does not exist:

```suggestion
        if (!resp.result || (resp.result.length === 0 || resp.result[0] === '' || resp.result[0] === '[]')) {
```

libV2/tasks/CreateSnapshot.js
```diff
@@ -55,6 +73,25 @@ class CreateSnapshot extends BaseTask {
...
+        const resp = await this.withWarp10(async warp10 => warp10.fetch({
+            className: 'utapi.snapshot.master',
+            labels: {
+                node: warp10.nodeId,
+            },
+            start: 'now',
+            stop: -1,
+        }));
+
+        if (resp.result && (resp.result.length === 0 || resp.result[0] === '' || resp.result[0] === '[]')) {
```
jonathan-gramain: same here

jonathan-gramain (approved):

alexanderchan-scality (approved):

PR #1251, opened 2022-05-26
ft(UTAPI-76): Breakout disk usage for leveldb and datalog

tmacro:


libV2/tasks/DiskUsage.js
```diff
@@ -249,16 +267,19 @@ class MonitorDiskUsage extends BaseTask {
...
             moduleLogger.error(`error calculating disk usage for ${this._path}`, { error });
             return;
         }
 
-        this._pushMetrics({ diskUsage: size });
+        this._pushMetrics({ leveldbBytes, datalogBytes });
 
+        const size = leveldbBytes + datalogBytes;
         if (this._hardLimit !== null) {
             moduleLogger.info(`warp 10 leveldb using ${formatDiskSize(size)} of disk space`, { usage: size });
```
jonathan-gramain:
```suggestion
            moduleLogger.info(`warp 10 using ${formatDiskSize(size)} of disk space`, { leveldbBytes, datalogBytes });
```

tests/functional/v2/task/testDiskUsage.js
```diff
@@ -56,12 +59,23 @@ describe('Test MonitorDiskUsage', () => {
...
+                const expectedSingleSize = emptyDirSize + testCase.expected + (emptyFileSize * testCase.count);
+                const expectedTotalSize = expectedSingleSize * 2;
+                assert.strictEqual(task.usage, expectedTotalSize);
+                // Should equal the usage minus the empty datalog
+                await assertMetricValue('utapi_monitor_disk_usage_leveldb_bytes', expectedSingleSize);
+                await assertMetricValue('utapi_monitor_disk_usage_datalog_bytes', expectedSingleSize);
+            });
+    });
+
+    it('should fail if a subpath does not exists', () => {
```
jonathan-gramain:
```suggestion
    it('should fail if a subpath does not exist', () => {
```

jonathan-gramain (approved):

alexanderchan-scality (approved):

libV2/tasks/DiskUsage.js
```diff
@@ -129,9 +143,13 @@ class MonitorDiskUsage extends BaseTask {
         return this._program.lock !== undefined;
     }
 
-    _getUsage() {
-        moduleLogger.debug(`calculating disk usage for ${this._path}`);
-        return getFolderSize(this._path);
+    // eslint-disable-next-line class-methods-use-this
+    async _getUsage(path) {
+        moduleLogger.debug(`calculating disk usage for ${path}`);
+        if (!fs.existsSync(path)) {
```
ghost, line 149: Hello! LGTM, I was just curious why the sync method is used in an async function? :slightly_smiling_face:

ghost (approved):

libV2/tasks/DiskUsage.js
tmacro: Mostly because I didn't want to wrap the callback API in a promise, but I see that the fs.promise API is no longer experimental so I can use that instead

tmacro: Actually as I read the docs more it seems the callback version is deprecated ([docs](https://nodejs.org/docs/latest-v16.x/api/fs.html#fsexistspath-callback)) and the promise version doesn't exist. `getFolderSize` is calling `stat` under the hood like the docs suggest so I think the sync usage is fine here.

PR #1258, opened 2022-07-21
bf(UTAPI-78): Fix second stage user auth with no resources

tmacro:


libV2/server/middleware.js
```diff
@@ -137,7 +137,7 @@ async function authV4Middleware(request, response, params) {
     } catch (error) {
         request.logger.error('error during authentication', { error });
         // rethrow any access denied errors
-        if (error.is && error.is.AccessDenied) {
+        if ((error.is && error.is.AccessDenied) || (error.utapiError && error.AccessDenied)) {
```
naren-rajendran, line 140: How can error be both `utapiError` and `AccessDenied`?

jonathan-gramain (approved):

libV2/server/middleware.js
tmacro: Utapi uses its own errors that have the `utapiError` property set so it can tell the difference in dependency errors and ones that it threw itself. One way its used is in development mode where it will let anything through in the response vs hiding it with an Internal Error in production https://github.com/scality/utapi/blob/ff5a75bb11a14530f93cea3e3884d74e1aef8cea/libV2/server/middleware.js#L93

watsaqat (approved):

PR #1268, opened 2022-09-22
ft(UTAPI-71): Add nodesvc based image and release workflow

tmacro:


.github/workflows/build-ci.yaml
```diff
@@ -0,0 +1,77 @@
...
+  workflow_call:
+    secrets:
+      REGISTRY_LOGIN:
+        required: true
+      REGISTRY_PASSWORD:
+        required: true
+
+jobs:
+  warp10-ci:
+    uses: scality/workflows/.github/workflows/docker-build.yaml@1.0.2
```
jonathan-gramain: Just wondering, is there a reason to use specifically 1.0.2 here instead of v1?

.github/workflows/tests.yaml
```diff
@@ -193,8 +134,112 @@ jobs:
...
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip
+    - name: Install python deps
+      run: |
+        pip install requests
+        pip install redis
+    - name: install dependencies
+      run: yarn install --frozen-lockfile --network-concurrency 1
+    - name: Wait for warp10 a little bit
```
jonathan-gramain: May help debugging to be more explicit

```suggestion
    - name: Wait for warp10 for 60 seconds
```

jonathan-gramain (approved):

watsaqat (approved):

.github/workflows/build-ci.yaml
tmacro: I needed to add support for LFS, though it looks like the v1 tag has been retagged now so I'll change it over

PR #1270, opened 2022-10-05
impr(UTAPI-83): Add warp 10 release workflow

tmacro:


naren-rajendran (approved):

.github/workflows/release-warp10.yaml
```diff
@@ -0,0 +1,41 @@
...
+  release:
+    if: ${{ inputs.create-github-release }}
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - uses: softprops/action-gh-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          name: Release scality/warp10:${{ github.event.inputs.tag }}
```
jonathan-gramain: This looks more accurate but please correct my suggestion if it is not correct:

```suggestion
          name: Release registry.scality.com/utapi/warp10:${{ github.event.inputs.tag }}-warp10
```

jonathan-gramain (approved):

Emylou-s (changes requested):
There seems to be significant changes between the dockerfile used in utapi repo and the one used in federation.
Among others one I can see is the line in Federation : 

`# Install protobuf extestion
ADD ./warp10-ext-protobuf-1.1.0-uberjar.jar /opt/warp10/lib/
`

How can we make sure the generated image will be compatible with what's built in Federation.

I have the feeling that there will be significant work needed in Federation to match this new image.

tmacro:
The image is compatible with the one in federation. The only difference aside from the bump to the protobuf extension is the
inclusion of the sensision prometheus exporter which is disabled by default.

Emylou-s (approved):

PR #1271, opened 2022-10-13
bf(UTAPI-85): Allow host and port to be overridden

tmacro:


jonathan-gramain (approved):

Emylou-s (approved):

Emylou-s:
@tmacro could we eventually use this PR to add --no-cache to the apk add stuff ?

tmacro:
@Emylou-s since the `apk install` is run the the `builder` stage we don't need to worry about cleaning up. The only thing copied over is the `warp10_sensision_exporter` binary

PR #1298, opened 2023-11-16
impr(UTAPI-97): Add config option to reindex only latest version in object locked buckets

tmacro:


lib/UtapiReindex.js
```diff
@@ -61,6 +61,10 @@ class UtapiReindex {
             this._log = new werelogs.Logger('UtapiReindex', { level, dump });
         }
 
+        if (config && config.onlyCountLatestWhenObjectLocked) {
+            this._onlyCountLatestWhenObjectLocked = true;
```
jonathan-gramain: To be clean we should initialize this field to `false` when not set.

lib/reindex/s3_bucketd.py
```diff
@@ -135,6 +142,26 @@ def _list_bucket(self, bucket, **kwargs):
...
+        try:
+            resp = self._do_req(url)
+            if resp.status_code == 200:
+                return resp.json()
+            else:
+                _log.error('Error getting bucket attributes bucket:%s status_code:%s'%(name, resp.status_code))
+                raise InvalidListing(name)
+        except ValueError as e:
+            _log.exception(e)
+            _log.error('Invalid attributes response body! bucket:%s'%name)
```
jonathan-gramain, line 155: Do we return `undefined` on purpose here? I.e. should we `raise` instead?

lib/reindex/s3_bucketd.py
```diff
@@ -49,7 +51,7 @@ def inner(*args, **kwargs):
         return urllib.parse.quote(val.encode('utf-8'))
     return inner
 
-Bucket = namedtuple('Bucket', ['userid', 'name'])
+Bucket = namedtuple('Bucket', ['userid', 'name', 'object_locked'], defaults=[False])
```
jonathan-gramain: I was a bit confused, I suggest to rename `object_locked` to `object_lock_enabled`

lib/reindex/s3_bucketd.py
```diff
@@ -216,6 +248,12 @@ def _sum_objects(self, bucket, listing):
                     # bucketd v7 returns an encoded string
                     data = json.loads(obj['value'])
                     size = data.get('content-length', 0)
+
+                if self._only_latest_when_locked and bucket.object_locked and '\x00' in obj['key']:
```
jonathan-gramain: I have the feeling checking for a zero-byte is too low-level. We should instead either rely on a `DelimiterVersions` listing for all cases, or change the listing type to `DelimiterMaster` in this particular case. I don't know how easy or hard it would be to change this.

For example in V1 metadata, latest delete markers would only have a versioned key and no master key.

jonathan-gramain (changes requested):

lib/reindex/s3_bucketd.py
tmacro: That was my original feeling as well, but the feature is only requested for v0. I'm hesitant to do too much work on this as utapi is essentially "finished" and only receiving bug fixes (and small customer driven features like this one). Using a different listing type could be possible, but I can't remember the differences off the top of my head.

jonathan-gramain: I get your point, but I don't think we can "trust" customers to never ever use utapi v1 with v1 buckets. Don't forget that v1 buckets are in the corner and already available in 9.2.0.4.

This being said I have no reason to block the PR if we say that utapi v1 is not compatible with metadata v1, but we need to be careful here and we should probably ask @rahulreddy what he thinks from product perspective.

lib/reindex/s3_bucketd.py
```diff
@@ -49,7 +51,7 @@ def inner(*args, **kwargs):
         return urllib.parse.quote(val.encode('utf-8'))
     return inner
 
-Bucket = namedtuple('Bucket', ['userid', 'name'])
+Bucket = namedtuple('Bucket', ['userid', 'name', 'object_locked_enabled'], defaults=[False])
```
jonathan-gramain: May be better as `object_lock_enabled`?

lib/UtapiReindex.js
```diff
@@ -61,6 +61,8 @@ class UtapiReindex {
             this._log = new werelogs.Logger('UtapiReindex', { level, dump });
         }
 
+        this._onlyCountLatestWhenObjectLocked = (config && config.onlyCountLatestWhenObjectLocked !== undefined);
```
jonathan-gramain: This looks strange to me, shouldn't we not activate the option if `config.onlyCountLatestWhenObjectLocked` is `false`?

lib/reindex/s3_bucketd.py
```diff
@@ -199,15 +235,14 @@ def get_next_upload_id(p):
     def _sum_objects(self, bucket, listing):
         count = 0
         total_size = 0
-        last_master = None
-        last_size = None
+        last_key = None
         for status_code, payload in listing:
-            contents = payload['Contents'] if isinstance(payload, dict) else payload
+            pprint(payload)
```
jonathan-gramain: remove this?

lib/reindex/s3_bucketd.py
```diff
@@ -216,47 +251,49 @@ def _sum_objects(self, bucket, listing):
...
+            if p is None:
+                return ''
+            return p.get('NextVersionIdMarker', '')
 
         params = {
-            'listingType': 'Basic',
-            'maxKeys': 1000,
-            'gt': get_next_marker,
+            'listingType': 'DelimiterVersions',
+            'maxKeys': 2,
```
jonathan-gramain: I believe this should be `1000`

jonathan-gramain (changes requested):

tests/utils/mock/BucketD.js
```diff
@@ -112,6 +112,15 @@ class BucketD {
         return body;
     }
 
+    _getBucketVersionsResponse(bucketName) {
+        const body = {
+            CommonPrefixes: [],
+            IsTruncated: false,
+            Versions: this._bucketContent[bucketName] || [],
```
jonathan-gramain: It probably doesn't matter in the test, but technically to be mocking the same response we should add a `"versionId": "null"` field to each entry.

jonathan-gramain (approved):

tests/utils/mock/BucketD.js
```diff
@@ -112,6 +112,17 @@ class BucketD {
         return body;
     }
 
+    _getBucketVersionResponse(bucketName) {
+        const body = {
+            CommonPrefixes: [],
+            IsTruncated: false,
+            Versions: (this._bucketContent[bucketName] || [])
+                // patch in a versionId to more closely match the real response
+                .map(entry => ({ ...entry, versionId: "null" })),
```
fredmnl: Shouldn't we fix this linting error? `"null" -> 'null'`

lib/UtapiReindex.js
```diff
@@ -61,6 +61,8 @@ class UtapiReindex {
             this._log = new werelogs.Logger('UtapiReindex', { level, dump });
         }
 
+        this._onlyCountLatestWhenObjectLocked = (config && config.onlyCountLatestWhenObjectLocked === true);
```
fredmnl, line 64: Question from a Node noob: would `(config?.onlyCountLatestWhenObjectLocked === true)` achieve the same thing?

fredmnl (approved):

lib/UtapiReindex.js
tmacro: Yeah it would give the same result. The reason I'm using the older `(config && config...)` is that the linter version used doesn't support the `?` syntax.

PR #1304, opened 2024-06-06
Improvement/utapi 103/support reindex by account

tmacro:
This PR adds support for limiting reindex to particular account(s).
As part of the changes, rather than having `--bucket` and `--account` differ in behavior, and to have a more consistent CLI, I've pulled in some of the requested features from S3C-8077 (support multiple buckets).
I also added a `--dry-run` flag to skip redis updates to help both dev and field use.

Modifies:
`--bucket`: Can now be passed multiple times to reindex multiple buckets

Adds:
`--account`: Limit reindex to an account canonical Id. Can be passed multiple times.
`--account-file`: Read canonical Ids from a file. 1 per line.
`--bucket-file`: Read bucket names from a file. 1 per line.
`--dry-run`: Skip updating redis.

### DIFF
--- a/lib/reindex/s3_bucketd.py
+++ b/lib/reindex/s3_bucketd.py
@@ -168,6 +168,14 @@ class BucketDClient:
             _log.error('Unhandled exception getting bucket attributes bucket:%s'%name)
             raise
 
+    def get_bucket_md(self, name):
+        md = self._get_bucket_attributes(name)
+        canonId = md.get('owner')
+        if canonId is None:
+            _log.error('No owner found for bucket %s'%name)
+            raise InvalidListing(name)
+        return Bucket(canonId, name, md.get('objectLockEnabled', False))
+
     def list_buckets(self, name = None):
 
         def get_next_marker(p):
--- a/lib/reindex/s3_bucketd.py
+++ b/lib/reindex/s3_bucketd.py
@@ -63,6 +63,10 @@ class InvalidListing(Exception):
     def __init__(self, bucket):
         super().__init__('Invalid contents found while listing bucket %s'%bucket)
 
+class BucketNotFound(Exception):
+    def __init__(self, bucket):
+        super().__init__('Bucket %s not found'%bucket)
+
 class BucketDClient:
 
     '''Performs Listing calls against bucketd'''
@@ -151,7 +155,7 @@ class BucketDClient:
                 return resp.json()
             else:
                 _log.error('Error getting bucket attributes bucket:%s status_code:%s'%(name, resp.status_code))
-                raise InvalidListing(name)
+                raise BucketNotFound(name)
         except ValueError as e:
             _log.exception(e)
             _log.error('Invalid attributes response body! bucket:%s'%name)
--- a/lib/reindex/s3_bucketd.py
+++ b/lib/reindex/s3_bucketd.py
@@ -1,5 +1,6 @@
 import argparse
 import concurrent.futures as futures
+import functools
 import itertools
 import json
 import logging
@@ -141,6 +142,7 @@ class BucketDClient:
             else:
                 is_truncated = len(payload) > 0
 
+    @functools.lru_cache(maxsize=16)
     def _get_bucket_attributes(self, name):
         url = self.__url_attribute_format.format(addr=self._bucketd_addr, bucket=name)
         try:
--- a/lib/reindex/s3_bucketd.py
+++ b/lib/reindex/s3_bucketd.py
@@ -296,7 +296,7 @@ class BucketDClient:
                 total_size += size
 
         except InvalidListing:
-            _log.error('Invalid contents in listing. bucket:%s status_code:%s'%(bucket.name, status_code))
+            _log.error('Invalid contents in listing. bucket:%s'%bucket.name)
             raise InvalidListing(bucket.name)
         return count, total_size
 
--- a/lib/reindex/s3_bucketd.py
+++ b/lib/reindex/s3_bucketd.py
@@ -38,6 +38,8 @@ def get_options():
     parser.add_argument("--debug", action='store_true', help="Enable debug logging")
     parser.add_argument("--dry-run", action="store_true", help="Do not update redis")
     group = parser.add_mutually_exclusive_group()
+    group.add_argument("-a", "--account", default=[], help="account canonical ID (all account buckets will be processed)", action="append", type=nonempty_string('account'))
+    group.add_argument("--account-file", default=None, help="file containing account canonical IDs, one ID per line", type=existing_file)
     group.add_argument("-b", "--bucket", default=[], help="bucket name", action="append", type=nonempty_string('bucket'))
     group.add_argument("--bucket-file", default=None, help="file containing bucket names, one bucket name per line", type=existing_file)
 
@@ -45,6 +47,9 @@ def get_options():
     if options.bucket_file:
         with open(options.bucket_file) as f:
             options.bucket = [line.strip() for line in f if line.strip()]
+    elif options.account_file:
+        with open(options.account_file) as f:
+            options.account = [line.strip() for line in f if line.strip()]
 
     return options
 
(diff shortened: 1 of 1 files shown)


lib/reindex/s3_bucketd.py
```diff
@@ -443,11 +444,19 @@ def log_report(resource, name, obj_count, total_size):
...
-            pipeline = redis_client.pipeline(transaction=False)  # No transaction to reduce redis load
-            for bucket, report in bucket_reports.items():
-                update_redis(pipeline, 'buckets', bucket, report['obj_count'], report['total_size'])
-                log_report('buckets', bucket, report['obj_count'], report['total_size'])
-            pipeline.execute()
+            if options.dry_run:
+                for bucket, report in bucket_reports.items():
+                    _log.warning(
+                        "DryRun: resource buckets [%s] will be not updated with obj_count %i and total_size %i" % (
+                            bucket, report['obj_count'], report['total_size']
```
jonathan-gramain: I would suggest `_log.info`, as it is a dry-run it is expected by the user.

Also thinking the formulation is somewhat odd or sounds negative and could be improved but I don't have a good substitution in mind at the moment. Thinking that the message should just convey what are the values that would be updated could work better (maybe just rewording as `would be updated` could work).

lib/reindex/s3_bucketd.py
```diff
@@ -459,24 +468,35 @@ def log_report(resource, name, obj_count, total_size):
...
-        for chunk in chunks(without_failed, ACCOUNT_UPDATE_CHUNKSIZE):
-            pipeline = redis_client.pipeline(transaction=False) # No transaction to reduce redis load
-            for userid, report in chunk:
-                update_redis(pipeline, 'accounts', userid, report['obj_count'], report['total_size'])
-                log_report('accounts', userid, report['obj_count'], report['total_size'])
-            pipeline.execute()
+        if options.dry_run:
+            for userid, report in account_reports.items():
+                _log.warning(
+                    "DryRun: resource account [%s] will be not updated with obj_count %i and total_size %i" % (
```
jonathan-gramain: same comment here for the messaging and `_log.info`

lib/reindex/s3_bucketd.py
```diff
@@ -32,12 +33,33 @@ def get_options():
...
+    def inner(value):
+        if not value.strip():
+            raise argparse.ArgumentTypeError("%s: value must not be empty"%flag)
+        return value
+    return inner
+
+def existing_file(path):
+    path = Path(path).resolve()
+    if not path.exists():
+        raise argparse.ArgumentTypeError("File does not exist")
```
jonathan-gramain: Could be nice to show the path to the file (may not be obvious to the user which file doesn't exist)

lib/reindex/s3_bucketd.py
```diff
@@ -204,9 +229,11 @@ def get_next_marker(p):
 
             if buckets:
                 yield buckets
-                if name is not None:
-                    # Break on the first matching bucket if a name is given
-                    break
+                if names:
+                    seen_buckets.update(b.name for b in buckets)
```
jonathan-gramain: I would rather do the `update` with the bucket name just processed after `buckets.append(bucket)`

lib/reindex/s3_bucketd.py
```diff
@@ -204,9 +229,11 @@ def get_next_marker(p):
...
+                    # Break if we have seen all the buckets we are looking for
+                    if all(b in seen_buckets for b in names):
+                        break
```
jonathan-gramain: Minor suggestion: if we know that `names` doesn't have duplicate bucket names (or alternatively we could ensure it doesn't), it should be enough to check that the size of the set is equal to `len(names)`.

lib/reindex/s3_bucketd.py
```diff
@@ -504,10 +518,12 @@ def log_report(resource, name, obj_count, total_size):
 
     stale_buckets = set()
     recorded_buckets = set(get_resources_from_redis(redis_client, 'buckets'))
-    if not options.bucket:
+    if not options.bucket and not options.account:
         stale_buckets = recorded_buckets.difference(observed_buckets)
     elif options.bucket:
         stale_buckets = { b for b in options.bucket if b not in observed_buckets }
+    elif options.account:
+        _log.warning('Stale buckets will not be cleared when using the --account or --account-file flags')
```
jonathan-gramain: It might be slightly reorganized for simplicity as

```
if options.bucket:
    ...
elif options.account:
    ...
else:
    # neither bucket nor account
    ...
```

lib/reindex/s3_bucketd.py
```diff
@@ -542,12 +558,21 @@ def log_report(resource, name, obj_count, total_size):
                     log_report('accounts', userid, report['obj_count'], report['total_size'])
                 pipeline.execute()
 
+        if options.account:
+            for account in options.account:
+                if account in failed_accounts:
+                    _log.error("No metrics updated for %s, one or more buckets failed" % account)
```
jonathan-gramain:
```suggestion
                    _log.error("No metrics updated for account %s, one or more buckets failed" % account)
```

jonathan-gramain (approved):
LGTM with some minor suggestions

lib/reindex/s3_bucketd.py
dvasilas: Just another proposition:

`DryRun: obj_count %i and total_size %i was calculated for resource bucket [%s]. The bucket has not been updated.`

lib/reindex/s3_bucketd.py
```diff
@@ -32,12 +33,33 @@ def get_options():
...
     parser.add_argument("-w", "--worker", default=10, type=int, help="Number of workers")
-    parser.add_argument("-b", "--bucket", default=None, help="Bucket to be processed")
     parser.add_argument("-r", "--max-retries", default=2, type=int, help="Max retries before failing a bucketd request")
     parser.add_argument("--only-latest-when-locked", action='store_true', help="Only index the latest version of a key when the bucket has a default object lock policy")
     parser.add_argument("--debug", action='store_true', help="Enable debug logging")
     parser.add_argument("--dry-run", action="store_true", help="Do not update redis")
-    return parser.parse_args()
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument("-b", "--bucket", default=[], help="bucket name", action="append", type=nonempty_string('bucket'))
+    group.add_argument("--bucket-file", default=None, help="file containing bucket names", type=existing_file)
```
dvasilas:
```suggestion
    group.add_argument("--bucket-file", default=None, help="file containing bucket names, one bucket name per line", type=existing_file)
```

lib/reindex/s3_bucketd.py
```diff
@@ -31,11 +33,38 @@ def get_options():
...
     parser.add_argument("-w", "--worker", default=10, type=int, help="Number of workers")
-    parser.add_argument("-b", "--bucket", default=None, help="Bucket to be processed")
     parser.add_argument("-r", "--max-retries", default=2, type=int, help="Max retries before failing a bucketd request")
     parser.add_argument("--only-latest-when-locked", action='store_true', help="Only index the latest version of a key when the bucket has a default object lock policy")
     parser.add_argument("--debug", action='store_true', help="Enable debug logging")
-    return parser.parse_args()
+    parser.add_argument("--dry-run", action="store_true", help="Do not update redis")
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument("-a", "--account", default=[], help="account canonical ID (all account buckets will be processed)", action="append", type=nonempty_string('account'))
+    group.add_argument("--account-file", default=None, help="file containing account canonical IDs", type=existing_file)
```
dvasilas: Same comment about giving a hint for the file format.

dvasilas (approved):

lib/reindex/s3_bucketd.py
tmacro: While fiddling with this I realized that with the addition of `get_bucket_md` the `names` param isn't used anywhere. I've simplified the function to just remove it.

PR #1305, opened 2024-06-25
bf: UTAPI-105 UtapiReindex: use list of redis sentinels

jonathan-gramain:
Use a list of Redis sentinels that are running on stateful nodes only, instead of localhost.

Previously, a stateless-only node wouldn't have a local sentinel no running, causing UtapiReindex to fail.

Added a failover mechanism in case of connection error on the current sentinel, to try each other one in turn.


jonathan-gramain:
ping

BourgoisMickael (approved):

nicolas2bert (approved):

PR #1311, opened 2025-04-02
UTAPI-110 update nodesvc-base image with node 22

tcarmet:


images/nodesvc-base/Dockerfile
```diff
@@ -1,20 +1,43 @@
-FROM ghcr.io/scality/federation/nodesvc-base:7.10.5.0
-
-ENV UTAPI_CONFIG_FILE=${CONF_DIR}/config.json
+ARG DEBIAN_FRONTEND=noninteractive
+ARG UTAPI_VERSION=latest
+FROM ghcr.io/scality/utapi:${UTAPI_VERSION} AS base
```
BourgoisMickael, line 3: So we really need 2 images for utapi ? Can we use only 1 image ? The other one doesn't use the same node version

BourgoisMickael (approved):

images/nodesvc-base/Dockerfile
tcarmet: I think we need two, because UTAPI is used on both ARTESCA and RING:
- Artesca uses the default image and does not use `supervisord` to starts its container
- I don't think they use `ballot` on artesca side
- They also don't have this requirement of having CONF_DIR DATA_DIR and all that stuff.

I can match the node version and image though.

tcarmet: Also I might have misunderstood what you meant, did you meant to use only the `FROM ghcr.io/scality/utapi:${UTAPI_VERSION}` and install and configure the necessary thing on top of it instead of using a multi stage build?

BourgoisMickael: No, multi stage is better we've got things we don't need in the other image.

But can we keep the same nodejs version in every repo for the final stage.

Maybe if we can't touch the version for the artesca one we can have a local build stage if it's only to `yarn install` we don't need to be dependant on the other image

images/nodesvc-base/Dockerfile
```diff
@@ -1,20 +1,43 @@
...
+
+FROM node:22.11.0-bookworm-slim
```
BourgoisMickael: We should be able to use `node:22.14` in this repo as well, even for artesca and the CI, I don't see any breaking or big changes in the changelog

images/nodesvc-base/Dockerfile
tcarmet: I have aligned the two images with the same base image.

images/nodesvc-base/Dockerfile
tcarmet: using `node:22.14.0-bookworm-slim` now, on both images.

BourgoisMickael (approved):

tmacro (approved):

anurag4DSB (approved):

PR #1312, opened 2025-04-08
UTAPI-110 outdated CI runner OS

tcarmet:
Update the CI runner to `ubuntu-22.04` as `ubuntu-20.04` is no longer supported.

### DIFF
--- a/.github/workflows/build-ci.yaml
+++ b/.github/workflows/build-ci.yaml
@@ -39,7 +39,7 @@ jobs:
         REDIS_IMAGE=ghcr.io/${{ github.repository }}/redis-ci:${{ github.sha }}
 
   vault-ci:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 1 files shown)


anurag4DSB (approved):

.github/workflows/build-ci.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
         REDIS_IMAGE=ghcr.io/${{ github.repository }}/redis-ci:${{ github.sha }}
 
   vault-ci:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
anurag4DSB, line 42: why not v24? (curious if there is a specific reason)

tcarmet: Playing it safe, I know that there's a bunch of changes in ubuntu `24.04` and due to the situation we are right now, I rather we update it in two steps.

tmacro (approved):

PR #1318, opened 2025-08-27
UTAPI-116: Check stderr and fix deprecation warnings

BourgoisMickael:
It floods the stderr logs of cloudserver and utapi

```
(node:4634) [DEP0174] DeprecationWarning: Calling promisify on a function that returns a Promise is likely a mistake.
```

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -21,7 +21,7 @@ on:
 
 env:
   AWS_SDK_JS_SUPPRESS_MAINTENANCE_MODE_MESSAGE: '1'
-  NODE_OPTIONS: '--trace-warnings'
+  # NODE_OPTIONS: '--trace-warnings'
 
 jobs:
   build-ci:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -21,6 +21,7 @@ on:
 
 env:
   AWS_SDK_JS_SUPPRESS_MAINTENANCE_MODE_MESSAGE: '1'
+  NODE_OPTIONS: '--trace-warnings'
 
 jobs:
   build-ci:
@@ -141,6 +142,11 @@ jobs:
     - name: ${{ matrix.test.name }}
       run: ${{ matrix.test.command }}
       env: ${{ matrix.test.env }}
+    - name: Print stderr logs
+      run: grep -H "" *.stderr.log || true
+      if: always()
+    - name: Error if stderr logs are not empty
+      run: if [ -n "$(find . -name '*.stderr.log' -size +0c)" ]; then exit 1; fi
 
   tests-v2-with-vault:
     needs:
@@ -246,6 +252,13 @@ jobs:
         tmate-server-rsa-fingerprint: ${{ secrets.TMATE_SERVER_RSA_FINGERPRINT }}
         tmate-server-ed25519-fingerprint: ${{ secrets.TMATE_SERVER_ED25519_FINGERPRINT }}
       if: ${{ ( github.event.inputs.debug == true || github.event.inputs.debug == 'true' ) }}
+    - name: Print stderr logs
+      run: grep -H "" *.stderr.log || true
+      if: always()
+    - name: Error if stderr logs are not empty
+      run: if [ -n "$(find . -name '*.stderr.log' -size +0c)" ]; then exit 1; fi
+
+
   tests-v2-without-sensision:
     needs:
       - build-ci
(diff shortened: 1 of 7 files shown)


.github/workflows/tests.yaml
```diff
@@ -18,6 +18,11 @@ on:
         required: false
         description: Timeout for ssh connection to worker (minutes)
         default: 30
+
+env:
+  AWS_SDK_JS_SUPPRESS_MAINTENANCE_MODE_MESSAGE: '1'
+  # NODE_OPTIONS: '--trace-warnings'
```
tcarmet, line 24:
```suggestion
```

.github/workflows/tests.yaml
```diff
@@ -137,6 +142,10 @@ jobs:
     - name: ${{ matrix.test.name }}
       run: ${{ matrix.test.command }}
       env: ${{ matrix.test.env }}
+    - name: Print stderr logs
+      run: grep -H "" *.stderr.log || true
```
tcarmet, line 146: Is this like for debugging purposes? If so let's make sure it is run everytime.
```suggestion
      run: grep -H "" *.stderr.log || true
      if: always()
```

.github/workflows/tests.yaml
```diff
@@ -242,6 +251,12 @@ jobs:
         tmate-server-rsa-fingerprint: ${{ secrets.TMATE_SERVER_RSA_FINGERPRINT }}
         tmate-server-ed25519-fingerprint: ${{ secrets.TMATE_SERVER_ED25519_FINGERPRINT }}
       if: ${{ ( github.event.inputs.debug == true || github.event.inputs.debug == 'true' ) }}
+    - name: Print stderr logs
+      run: grep -H "" *.stderr.log || true
```
tcarmet, line 256:
```suggestion
      run: grep -H "" *.stderr.log || true
      if: always()
```

.github/workflows/tests.yaml
```diff
@@ -358,3 +373,7 @@ jobs:
...
```
tcarmet, line 379:
```suggestion
      run: grep -H "" *.stderr.log || true
      if: always()
```

tcarmet (approved):

.github/workflows/tests.yaml
```diff
@@ -137,6 +142,10 @@ jobs:
     - name: ${{ matrix.test.name }}
       run: ${{ matrix.test.command }}
       env: ${{ matrix.test.env }}
+    - name: Print stderr logs
+      run: grep -H "" *.stderr.log || true
+    - name: Error if stderr logs are not empty
```
jonathan-gramain, line 148: Maybe a bit extreme to fail the test? IDK, just a personal feeling, it's useful to have feedback on any new stderr message, maybe we can add a `continue-on-error: true` so it's visible but doesn't outright fail the test if the overall test suite is successful.

.github/workflows/tests.yaml
BourgoisMickael: I prefer to keep so we can easily uncomment to have stack trace if a warning appear. 

I commented because there is currently a warning ignored

.github/workflows/tests.yaml
BourgoisMickael: It's for debugging the warnings (from the step below). 

If there is a test error before there is no need to print those stderr, they already appear in regular CI output

libV2/redis.js
```diff
@@ -15,6 +15,49 @@ const moduleLogger = new LoggerContext({
...
 const CONNECTION_TIMEOUT = 30000;
 
+/**
+ * Promisifies a function. If the function already returns a promise,
+ * it returns it as is. Handles both callbacks and promises.
+ *
+ * @param {Function} originalFn The function to promisify.
+ * @returns {Function} A function that returns a promise.
+ */
+function flexiblePromisify(originalFn) {
```
jonathan-gramain: Is this a standard, proven function?
- if so, can you provide the ref
- if not, can you add unit tests to make sure it works as designed?

.github/workflows/tests.yaml
BourgoisMickael: That's the idea to catch immediately a warning, so we are not surprised later.

Then if we don't plan on fixing the warning, we can filter it out from the logs ([38badb7](https://github.com/scality/utapi/pull/1318/commits/38badb76a0402ff224b46f0b9b85bca94e9f3273))

A `continue-on-error: true` could be added at any moment if there are warnings that can't be fixed to pass the tests, but ideally I'd like to fail to make sure we notice it. 

And I'd like to have this in other components as well in long term

libV2/redis.js
BourgoisMickael: Nothing standard, made with gemini

here are unit tests: [7a12ad2](https://github.com/scality/utapi/pull/1318/commits/7a12ad252e1f37653096f67088a8311bfe4b7d4e)

jonathan-gramain (approved):

tmacro (approved):

PR #1332, opened 2026-04-15
UTAPI-122: Fix warp10 client timeouts

tmacro:
This PR reverts the upgrade of @senx/warp10 to 2.0.3
Versions newer than 1.0.14 contain a bug that prevents setting a custom request timeout.

### DIFF
--- a/libV2/warp10.js
+++ b/libV2/warp10.js
@@ -19,7 +19,7 @@ class Warp10Client {
         this._connectTimeout = (config && config.connectTimeout) || 30000;
         const host = (config && config.host) || '127.0.0.1';
         const port = (config && config.port) || 4802;
-        this._client = new Warp10().endpoint(`${proto}://${host}:${port}`, this._requestTimeout, this._connectTimeout);
+        this._client = new Warp10(`${proto}://${host}:${port}`, this._requestTimeout, this._connectTimeout);
     }
 
     async update(payload) {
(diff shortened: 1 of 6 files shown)


libV2/warp10.js
```diff
@@ -15,11 +15,11 @@ class Warp10Client {
...
         this.nodeId = (config && config.nodeId) || _config.nodeId;
         const proto = (config && config.tls) ? 'https' : 'http';
-        this._requestTimeout = (config && config.requestTimeout) || 30000;
-        this._connectTimeout = (config && config.connectTimeout) || 30000;
         const host = (config && config.host) || '127.0.0.1';
         const port = (config && config.port) || 4802;
-        this._client = new Warp10().endpoint(`${proto}://${host}:${port}`, this._requestTimeout, this._connectTimeout);
+        this._requestTimeout = (config && config.requestTimeout) || 60000;
+        this._connectTimeout = (config && config.connectTimeout) || 60000;
+        this._client = new Warp10(`${proto}://${host}:${port}`, this._requestTimeout, this._connectTimeout);
```
anurag4DSB, line 22: Could we add some comments for the reasoning of choosing this value please for the future us.

anurag4DSB (approved):

jonathan-gramain (approved):

BourgoisMickael (approved):

libV2/warp10.js
tmacro: It's just to match the default value that actually gets used from the config
https://github.com/scality/utapi/blob/45e0ef1010d2eadee585ba2444475f691be6ed30/libV2/config/defaults.json#L20

PR #1334, opened 2026-07-23
Bump arsenal to 8.5.6 to support V2 replication bucket metadata

maeldonn:
Align with the arsenal version cloudserver ships. Arsenal >= 8.4.7 writes
bucket replication metadata in the V2 (multi-destination CRR) format
without the deprecated top-level destination. Arsenal 8.2.28 asserted that
field unconditionally and crashed utapi when reading such buckets.

Issue: UTAPI-124

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "engines": {
     "node": ">=22"
   },
-  "version": "8.2.6",
+  "version": "8.3.0",
   "description": "API for tracking resource utilization and reporting metrics",
   "main": "index.js",
   "repository": {
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
   "dependencies": {
     "@hapi/joi": "^17.1.1",
     "@senx/warp10": "1.0.14",
-    "arsenal": "git+https://github.com/scality/Arsenal#8.2.28",
+    "arsenal": "git+https://github.com/scality/Arsenal#8.5.6",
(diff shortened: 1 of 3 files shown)


maeldonn:
@leif-scality I think we should target a new `development/8.3` branch for this. What do you think?

SylvainSenechal (approved):

SylvainSenechal:
> @leif-scality I think we should target a new `development/8.3` branch for this. What do you think?

ye can

benzekrimaha (approved):
LGTM, well done

leif-scality:
So this PR will make utapi not compatible with `cloudserver 9.3` and previous versions? Only `cloudserver 9.4` will work after this? Maybe we should do a `development/8.3`

@nicolas2bert

leif-scality:
> So this PR will make utapi not compatible with `cloudserver 9.3` and previous versions? Only `cloudserver 9.4` will work after this? Maybe we should do a `development/8.3`
> 
> @nicolas2bert

This fixed the integration problem
Federation build: https://github.com/scality/Federation/pull/6993
Green Integration: https://github.com/scality/Integration/actions/runs/30082259417

leif-scality (changes requested):
LGTM, but we need merge it in a new development/8.3

leif-scality (approved):

PR #1335, opened 2026-09-15
Migrate to AWS SDK v3

francoisferrand:
aws-sdk v2 is end-of-support and no longer gets updates, and every other Scality component has moved on. utapi only used it for the IAM client — in `bin/ensureServiceUser` and the test helpers — so this swaps it for `@aws-sdk/client-iam`. No S3 usage anywhere, so none of the usual v3 pain around path style or streaming bodies applies.

The not-found check is the part worth reviewing. v2's `err.code` is gone in v3, so `err.code === 'NoSuchEntity'` silently stops matching: on a fresh install `GetUser` throws instead of falling through to `CreateUser`, and the script exits 1. That is exactly how cloudserver broke S3C installs in CLDSRV-915, where the migration swapped it for `err.name === 'NoSuchEntity'` — also never true, since Vault's wire code deserializes to `NoSuchEntityException`. This matches against the modeled exception, like backbeat and cloudserver do today.

Two other shape changes: the client needs an explicit region now (taken from the usual env vars, defaulting to `us-east-1`, since Vault ignores it but SigV4 does not), and v3 leaves list fields undefined where v2 gave an empty array, so the `ListPolicies`/`ListUsers`/`ListAttachedUserPolicies` readers are guarded.

aws-sdk v2 still reaches us through `bucketclient → arsenal#8.2.4`, so `AWS_SDK_JS_SUPPRESS_MAINTENANCE_MODE_MESSAGE` has to stay in CI for now — dropping it would put the maintenance warning on stderr and trip the empty-stderr gate. BKTCLT-50 tracks moving that pin.

Targeting `development/8.3` since 8.3.0 has not shipped yet.

Verified with lint, the unit suite, and a mock Vault-style IAM endpoint driving `ensureServiceUser` end to end: fresh-install create, idempotent re-run, SigV4 credential scope, and both empty-list XML shapes. The functional suite needs a live Vault, so CI is the real check there.

Issue: UTAPI-115


benzekrimaha (approved):
LGTM

delthas (approved):

